PR #22, opened 2018-12-10
this branch should be used after salvatore will accept the PR

MeirShpilraien:


MeirShpilraien:
this should address the issue: https://github.com/RedisLabsModules/RedisGears/pull/119

PR #99, opened 2019-04-14
added network recovery mechanism

MeirShpilraien:


MeirShpilraien:
- [ ] add tests
- [ ] add shard failure recovery

MeirShpilraien:
@itamarhaber can you pls review this, I will leave the shard failure recovery to another PR.

src/cluster.c
```diff
@@ -571,20 +674,47 @@ int Cluster_GetClusterInfo(RedisModuleCtx *ctx, RedisModuleString **argv, int ar
         }else{
             RedisModule_ReplyWithStringBuffer(ctx, "None", strlen("None"));
         }
+        RedisModule_ReplyWithStringBuffer(ctx, "runid", strlen("runid"));
```
itamarhaber: Good thing that RG.GETCLUSTERINFO is not documented or you'd have had to update the docs because of this change in the reply :)

MeirShpilraien: Prefer not to document it yet, it might change and it is not something that I want users to use.

MeirShpilraien:
@itamarhaber fixed.

PR #102, opened 2019-04-16
Add timings to execution plan steps

itamarhaber:
This PR introduces a new command `RG.GETEXECUTION <execution-id>` (temporary name, perhaps should be merged with `DUMPEXECUTIONS`) that provides per-step information about the execution plan.

```text
127.0.0.1:6379> DEBUG POPULATE 10000
OK
127.0.0.1:6379> RG.PYEXECUTE "GB().map(lambda x: x['value']).run()" UNBLOCKING
"0000000000000000000000000000000000000000-0"
127.0.0.1:6379> RG.GETEXECUTION 0000000000000000000000000000000000000000-0
 1) "status"
 2) "done"
 3) "state"
 4) "ok"
 5) "results_len"
 6) (integer) 10000
 7) "total_duration"
 8) (integer) 33582773
 9) "steps"
10) 1) 1) "type"
       2) "Collect"
       3) "duration"
       4) (integer) 0
    2) 1) "type"
       2) "Map"
       3) "duration"
       4) (integer) 4986693
    3) 1) "type"
       2) "Map"
       3) "duration"
       4) (integer) 4967869
    4) 1) "type"
       2) "Map"
       3) "duration"
       4) (integer) 10971792
    5) 1) "type"
       2) "Reader"
       3) "duration"
       4) (integer) 9794417
```

TODO:
- [x] Add the function name and its arguments to each step
- [x] Separate errors from results in ExecutionPlan
- [x] Explain discrepancy between Total and actual sum of steps: execution overhead <- should be documented
- [x] Make it work in the cluster (i.e. have a gear collect and aggregate the steps?)
- [x] Render final duration output of `GETEXECUTION` &  `DUMPEXECUTIONS` to msec
- [x] Write tests
- [x] Document new config params and commands


src/execution_plan.c
```diff
@@ -327,37 +325,40 @@ static void ExecutionPlan_Distribute(ExecutionPlan* ep){
...
         }
         if(filterRes){
-            return record;
+            goto end;
         }else{
             RedisGears_FreeRecord(record);
         }
     }
-    return NULL;
+    return NULL;  // TODO: check if this is a reachable code path
```
MeirShpilraien: its reachable when there is no more records

src/execution_plan.c
```diff
@@ -509,24 +506,34 @@ static Record* ExecutionPlan_GroupNextRecord(ExecutionPlan* ep, ExecutionStep* s
         RedisGears_ListRecordAdd(listRecord, RedisGears_KeyRecordGetVal(record));
         RedisGears_KeyRecordSetVal(record, NULL);
         RedisGears_FreeRecord(record);
+        STOP_TIMER;
+    	step->executionDuration += DURATION;
```
MeirShpilraien: why not using the ADD_DURATION here?

src/execution_plan.c
```diff
@@ -1694,10 +1755,72 @@ int ExecutionPlan_ExecutionsDump(RedisModuleCtx *ctx, RedisModuleString **argv,
...
+	}
+
+	RedisModule_ReplyWithStringBuffer(ctx, "state", strlen("state"));
+	if(ep->isErrorOccure){
+		RedisModule_ReplyWithStringBuffer(ctx, "error", strlen("error"));
+		RedisModule_ReplyWithStringBuffer(ctx, "message", strlen("message"));
+        // TODO: replace with a full backtrace
+        assert(RedisGears_GetRecordsLen(ep) > 0);
+		Record* record = RedisGears_GetRecord(ep, 0);
+        assert(RedisGears_RecordGetType(record) == ERROR_RECORD);
```
MeirShpilraien: the asserts here are wrong, but I guess you fix it when you add the error records list.

src/execution_plan.c
```diff
@@ -1694,10 +1755,72 @@ int ExecutionPlan_ExecutionsDump(RedisModuleCtx *ctx, RedisModuleString **argv,
...
+
+	const char* id = RedisModule_StringPtrLen(argv[1], NULL);
+	ExecutionPlan* ep = RedisGears_GetExecution(id);
+
+	if(!ep){
+		RedisModule_ReplyWithError(ctx, "execution plan does not exist");
+		return REDISMODULE_OK;
+	}
+
+    pthread_mutex_lock(&epData.mutex);
```
MeirShpilraien: why you need to acquire this lock?

src/execution_plan.c
itamarhaber: Because I forgot... :)

src/execution_plan.c
itamarhaber: Ok, cool.

src/execution_plan.c
itamarhaber: Yep - this is just a placeholder.

src/execution_plan.c
itamarhaber: Hmmm - I assumed it is needed for the same reason it exists in DUMPEXECUTIONS. Is that not the case?

MeirShpilraien: Sorry you are right, now the problem that between the time you took the execution to the time locking it might be that the execution got deleted. We are gone need a lock that allow the same thread to lock it multiple times (like I did with the LockHandler). Let me sleep on a good solution for that..

itamarhaber:
TY.

I ran 100 executions, with and without timings. The average total duration is 36462353 vs. 40251351 so making this configurable is a good point <- added `ProfileExecutions` defaults to 0 (disabled)... now I need @rafie's config branch merged :)

src/execution_plan.c
itamarhaber: Resolved via  f227215

src/execution_plan.c
itamarhaber: Resolved via  f227215

src/execution_plan.c
```diff
@@ -16,10 +16,10 @@
...
-#define INIT_TIMER  struct timespec _ts,_te;
+#define INIT_TIMER  struct timespec _ts = {0}, _te = {0};
 #define GETTIME(t)  clock_gettime(CLOCK_REALTIME, t);
-#define START_TIMER GETTIME(&_ts);
-#define STOP_TIMER  GETTIME(&_te);
+#define START_TIMER if(GearsConfig_GetProfileExecutions()) GETTIME(&_ts);
+#define STOP_TIMER  if(GearsConfig_GetProfileExecutions()) GETTIME(&_te);
 #define DURATION    ((long long)1000000000 * (_te.tv_sec - _ts.tv_sec) \
                     + (_te.tv_nsec - _ts.tv_nsec))
 #define ADD_DURATION(d) STOP_TIMER; \
```
MeirShpilraien: shouldn't this also have the if(GearsConfig_GetProfileExecutions())

itamarhaber: Do you mean ADD_DURATION? I guess so.

itamarhaber: Resolved via 331f9aa

src/execution_plan.c
itamarhaber: This should be resolved via a97de6e

src/redisgears_python.c
```diff
@@ -30,6 +30,10 @@ static RedisModuleCtx* currentCtx = NULL;
 static bool blockingExecute = true;
```
MeirShpilraien: all those vars should be move to thread local, this might break if we create an execution from another thread while also creating from the main thread (like for example, an execution that creates execution inside it). We will do it in another PR, lets just add a todo here

src/redisgears_python.c
```diff
@@ -1088,6 +1092,22 @@ static int RedisGearsPy_Execut(RedisModuleCtx *ctx, RedisModuleString **argv, in
...
+    RedisModuleBlockedClient *bc = privateData;
+    RedisModuleCtx *rctx = RedisModule_GetThreadSafeContext(bc);
+    Command_ReturnResults(ep, rctx);
+    RedisModule_UnblockClient(bc, NULL);
+    RedisGears_DropExecution(ep);
+    RedisModule_FreeThreadSafeContext(rctx);
+}
+
+int RedisGearsPy_ExecuteReturnResultsOnly(RedisModuleCtx *ctx, RedisModuleString **argv, int argc){
+    doneFunction = onDoneResultsOnly;
```
MeirShpilraien: thought maybe to give this function the callback as an arg and implement the onDoneResultsOnly on the command.c, what do you think?

src/execution_plan.c
itamarhaber: This has been "resolved" with a TODO and a big warning

src/redisgears_python.c
itamarhaber: Sure.

src/redisgears_python.c
itamarhaber: That was my original thinking and understanding from you. However. I found it difficult for technical reasons (or own lack of knowledge) - lets slack on this.

itamarhaber:
^ Passes valgrind with flying colors ;) We need to add it to the CI for PRs (and nightly of course)

src/redisgears_python.c
itamarhaber: Resolved via a "TODO"

src/redisgears_python.c
itamarhaber: Resolved via 635f722

itamarhaber:
This should be merge-able.

docs/commands.md
```diff
@@ -74,6 +74,39 @@ List of executions with
 * executionId
 * status (running/done)
 
+## RG.GETEXECUTION
+```sql
+RG.DUMPEXECUTIONS execution-id [ SHARD | CLUSTER ]
```
MeirShpilraien: this should be RG.GETEXECUTION right?

src/execution_plan.c
```diff
@@ -1812,7 +1822,7 @@ int ExecutionPlan_ExecutionGet(RedisModuleCtx *ctx, RedisModuleString **argv, in
             ".collect()"
             ".flatmap(lambda x: [i for i in x])"
             ".run(convertToStr=False, collect=False)", eid);
-        int res = RedisGearsPy_ExecuteReturnResultsOnly(ctx, fargv, 2);
```
MeirShpilraien: you allocate the fargv with RedisModule_Calloc should be RG_ALLOC

src/execution_plan.c
```diff
@@ -1812,7 +1822,7 @@ int ExecutionPlan_ExecutionGet(RedisModuleCtx *ctx, RedisModuleString **argv, in
...
+        int res = RedisGearsPy_ExecuteWithCallback(ctx, fargv, 2, onDoneResultsOnly);
         RedisModule_Free(fargv[1]);
```
MeirShpilraien: The only thing I do not like here is that you have to create it fargv with size of 2 (cause the function assume it also contains the command name). Lets add a todo here and leave it like this for now.

docs/commands.md
itamarhaber: Yep - resolved via f8911ec

src/execution_plan.c
itamarhaber: Done in f8911ec (also fixed to use RG_FREE instead).

src/execution_plan.c
itamarhaber: Added TODO in f8911ec

itamarhaber:
@MeirShpilraien perhaps we squash before merge?

MeirShpilraien:
@itamarhaber yes, I think you can do it here on github with squash and megre.

PR #107, opened 2019-04-26
Adds OSX deps to README

itamarhaber:
And also removes dead code that breaks compilation w/ clang


itamarhaber:
Thanks for the review @MeirShpilraien - merging.

PR #114, opened 2019-04-29
change exit(1) to assert(false)

MeirShpilraien:


src/execution_plan.c
```diff
@@ -1017,12 +1027,17 @@ static void ExecutionPlan_Main(ExecutionPlan* ep){
...
 	if(isDone){
 	    if(Cluster_IsClusterMode()){
-            Cluster_SendMsgM(NULL, ExecutionPlan_NotifyExecutionDone, ep->id, EXECUTION_PLAN_ID_LEN);
-            if((Cluster_GetSize() - 1) == ep->totalShardsCompleted){ // no need to wait to myself
-                ExecutionPlan_DoExecutionDoneActions(ep, rctx);
-            }else{
-                ep->status = WAITING_FOR_CLUSTER_TO_COMPLETE;
-            }
+	        if(memcmp(ep->id, Cluster_GetMyId(), REDISMODULE_NODE_ID_LEN) == 0){
+	            if((Cluster_GetSize() - 1) == ep->totalShardsCompleted){ // wait for all the shards to complete the execution
```
itamarhaber: Not sure this will hold when the cluster size decreases unexpectedly during execution - perhaps worry about that in another day though.

MeirShpilraien: Yes we do not yet support topolegy changes and shard crashes during execution yet. We still need to face those problems.

PR #121, opened 2019-05-06
Fixes to pass make on fresh Ubuntu 18.04

itamarhaber:
@rafie @MeirShpilraien 

Also:

- [ ] there is a dependency on redis-server in pack.sh... this is problematic-ish
- [ ] creating /opt/redislabs reuires sudo
- [ ] tools (eg. RLTest) should be installed for the user

### DIFF
--- a/build/libevent/Makefile
+++ b/build/libevent/Makefile
@@ -19,7 +19,7 @@ export CFLAGS=-fPIC
 
 all: build
 
-build: __sep bindirs
+build: __sep bindirs source
 ifeq (,$(wildcard $(BUILD_DIR)/Makefile))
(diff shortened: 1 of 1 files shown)


rafie:
* dependency on redis-server in pack.sh: this is nothing new. ramp requires redis-server.
* creating /opt/redislabs reuires sudo: I think this is the way to go. we can avoid sudo by checking if /opt/redislab exists, and kindly asking the user to create it.
* tools (eg. RLTest, ramp) should be installed: this can be done by having a Pipenv file for build/test environment. This can reduce the impact on the developer workstation. However, from examining system-setup.py, we already installing quite a bit there. Maybe it is reasonable to have our internal tools be a part of a virtual environment after all.

MeirShpilraien:
@rafie @itamarhaber I think we can close this one right?

rafie:
Yep, fixed long ago.

PR #122, opened 2019-05-07
see #101, switch to python 3.7

MeirShpilraien:


MeirShpilraien:
@rafie / @itamarhaber 
Need review again, I fix the circle failures.

MeirShpilraien:
@rafie pushing so I can continue working, we need to make sure this is working correctly on MAC.

PR #124, opened 2019-05-13
Fixed build issues (mostly on Mac OS X)

rafie:
- system-setup.py and general automation problems on Mac:
	- installation as non-root user,
	- no python2 on Mac,
	- addded $HOME/Library/Python/2.7/bin to PATH,
	- no realpath() on Mac,
	- mktemp on Mac,
	- disabled annoying Python 2 warnings,
	- creation of /opt/redislabs on Mac.
- Build problems on Mac:
	- --whole-archive linker issue on Mac,
	- ignore DEBUG=1 for libevent,
- PyEnv issues:
	- enabled site packages.


MeirShpilraien:
@rafie can we fix the conflict.
@itamarhaber can you confirm that this fix the build issue you had on mac?

itamarhaber:
I ran `make setup`, `make fetch` and `make all` on my installation that was already "dirty" from previous attempts. It looks like it is working but to really verify I'll need to re-install the OS from scratch and then try :)

PR #131, opened 2019-05-20
small fixes, better error checks, when string conversion failed create a byte array

MeirShpilraien:


src/redisgears_python.c
```diff
@@ -760,17 +778,46 @@ static void getAllValues(PyObject *list, double** values){
 static PyObject* createTensorFromBlob(PyObject *cls, PyObject *args){
     verifyRedisAILoaded();
     PyObject* typeName = PyTuple_GetItem(args, 0);
+    if(!PyUnicode_Check(typeName)){
+        PyErr_SetString(GearsError, "type argument must be a string");
+        return NULL;
+    }
+    PyObject* pyBlob = PyTuple_GetItem(args, 2);
+    if(!PyByteArray_Check(pyBlob)){
+        PyErr_SetString(GearsError, "blob argument must be a byte array");
```
itamarhaber: ...must be a bytesarray

src/redisgears_python.c
```diff
@@ -780,26 +827,66 @@ static PyObject* createTensorFromBlob(PyObject *cls, PyObject *args){
...
 
     PyObject* values = PyTuple_GetItem(args, 2);
-    for(long long i = 0 ; i < PyList_Size(values) ; ++i){
-        PyObject* val = PyList_GetItem(values, i);
-        RedisAI_TensorSetValueFromDouble(t, i, PyFloat_AsDouble(val));
+    PyObject* valuesIter = PyObject_GetIter(pyDims);
+    PyObject* currValue = NULL;
+    size_t index = 0;
+    while((currValue = PyIter_Next(valuesIter)) != NULL){
+        if(!PyFloat_Check(currValue)){
```
itamarhaber: I think the type needs to be according to whatever's in typeName

src/redisgears_python.c
```diff
@@ -1594,7 +1706,12 @@ static Record* RedisGearsPy_ToPyRecordMapperInternal(Record *record, void* arg){
     switch(RedisGears_RecordGetType(record)){
     case STRING_RECORD:
         str = RedisGears_StringRecordGet(record, &len);
+        // try to first decode it as string, if fails create a byte array.
```
itamarhaber: Isn't there some checktype method instead of relying on failure?

MeirShpilraien: On our side we can not know what the byte array represents. I am not sure if Python expose an api to check if bytearray is a valid string or not.

build/cpython/Makefile
```diff
@@ -105,6 +105,7 @@ pyenv: /opt/redislabs $(CPYTHON_PREFIX) $(PYENV_DIR)
 
 $(CPYTHON_PREFIX): /opt/redislabs
 	$(SHOW)make -C $(BUILD_DIR) install 2>&1 >$(BUILD_DIR)/python-install.log
+	$(SHOW)objcopy --localize-symbol PyGILState_Ensure --localize-symbol PyGILState_Release $(BINDIR)/libpython3.7m.a
```
itamarhaber: I tried finding a variable that has the lib names instead of hardcoding. It exists in Makefile.defs and there's already a commented-out include for it at the top of this file.

Issue #148, opened 2019-08-16
Logging being printed to stdout rather than through logger


andresrinivasan:
``` txt
1:M 16 Aug 2019 22:28:52.549 * <rg> RedisGears version 0.3.1, git_sha=be5c5fcdf2abaabe5ff62155d9c38e0ecaa97575
1:M 16 Aug 2019 22:28:52.550 * <rg> PythonHomeDir:/opt/redislabs/lib/modules/python3/
1:M 16 Aug 2019 22:28:52.550 * <rg> MaxExecutions:1000
1:M 16 Aug 2019 22:28:52.551 * <rg> ProfileExecutions:0
1:M 16 Aug 2019 22:28:52.551 * <rg> PythonAttemptTraceback:1
could not initialize RedisAI_GetAPIVersion
*1:M 16 Aug 2019 22:28:52.552 # <rg> could not initialize RediAI api, running without AI support.*
could not initialize RediSearch_CheckApiVersionCompatibility
*1:M 16 Aug 2019 22:28:52.553 # <rg> could not initialize RediSearch api, running without Search support.*
1:M 16 Aug 2019 22:28:52.598 * <rg> Initializing Python environment with: exec(open('/opt/redislabs/lib/modules/python3/.venv/bin/activate_this.py').read(), {'__file__': '/opt/redislabs/lib/modules/python3/.venv/bin/activate_this.py'})
1:M 16 Aug 2019 22:28:52.677 * Module 'rg' loaded from /opt/redislabs/lib/modules/redisgears.so
```

The lines`could not initialize RedisAI_GetAPIVersion`, and `could not initialize RediSearch_CheckApiVersionCompatibility` are being printed to stdout rather than being logged.

andresrinivasan:
What do those warnings mean?

MeirShpilraien:
@andresrinivasan those are new capability that was introduce in redis v5.0.3 that allows modules to use each other capability with a direct function call. The warning say that redisearch and redisai are not loaded so you will not be able to use them. I agree it should be printed only to the log file and not to the stdout.

MeirShpilraien:
It is possible now to use Log function (inside the gears function) to write to the redis log file:
```
Log(logLevel, msg)
```

PR #151, opened 2019-08-18
Partial PR for #148

andresrinivasan:
Re #148, removed redundant printf.

Signed-off-by: Andre Srinivasan <andre@redislabs.com>


MeirShpilraien:
@andresrinivasan this h file is the api expose by redisearch and taken as is from redisearch. It was modified and change couple for times and we should just take the new version as is from redisearch instead of modifying it here locally.

andresrinivasan:
Sorry, didn't realize that. I will make the change in the RediSearch repo.

MeirShpilraien:
And then we need to take the updated file from redisearch ...

PR #155, opened 2019-09-18
Changed the following:

MeirShpilraien:
1. if setting MaxExecutions to 0, no execution will be deleted automatically.
2. Events to execution are sent via a linked list and not with libevent to prevent deadlock
3. Keys only reader is now implemented in C to increase performance, in addition, if exect match (no prefix)
   is requested then 'scan' command will not be used.

### DIFF
--- a/src/keys_reader.c
+++ b/src/keys_reader.c
@@ -177,22 +177,19 @@ static Record* KeysReader_ReadKey(RedisModuleCtx* rctx, KeysReaderCtx* readerCtx
         return NULL;
     }
 
+    char* keyCStr = RG_ALLOC(keyLen + 1);
+    memcpy(keyCStr, keyStr, keyLen);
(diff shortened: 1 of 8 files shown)


src/keys_reader.c
```diff
@@ -149,7 +167,42 @@ static Record* ValueToRecordMapper(RedisModuleCtx* rctx, Record* record, RedisMo
...
+    RedisModuleKey *keyHandler = RedisModule_OpenKey(rctx, key, REDISMODULE_READ);
+    if(!keyHandler){
+        return NULL;
+    }
+
+    if(readerCtx->readValue){
+
+        record = RedisGears_KeyRecordCreate();
+
+        char* keyCStr = RG_ALLOC(keyLen + 1);
```
rafie: Possible to move outside the condition block.

PR #181, opened 2019-12-18
WIP: write_behind for Oracle and setup scripts

rafie:


recipes/write_behind/README.md
```diff
@@ -0,0 +1,99 @@
+# RedisGears Write-Behind Recipe
+
+## System requirements
+
+* Redis v5.0.7 or above,
+* Relational database: MySQL (any version will do) or Oracle (tested with 11g and 12c) database,
+* Redis Enterprise Software v5.4.11-2 and above,
```
MeirShpilraien: This is a requirement only if you use RedisEnterprise

recipes/write_behind/README.md
```diff
@@ -0,0 +1,99 @@
...
+* RedisGears module build for Ubuntu Bionic
+
+## Configuration
+
+TBD: key names and tables
+
+## Installing on Redis Enterprise Software host(s)
```
MeirShpilraien: You do not need RedisEnterprise to run it, you can run it with vanila Redis, So I think we do not need this part..

recipes/write_behind/oracle/rs/start-gear
```diff
@@ -1,6 +1,4 @@
 #!/bin/bash
 
-recipe=/opt/redislabs/lib/modules/gears_write_behind
-
-cd $recipe
+cd /opt/gears-wb
```
MeirShpilraien: Are you putting it there somewhere on the make part? We will once again will have permissions issue

recipes/pack.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
```
MeirShpilraien: What is this for?

rafie:
Didn't yet implement the MySQL/Oracle detection. It is coming.
The README describes how to integrate it with Redis Enterprise, when you use it. It's not a requirement but a constraint. I'm adding the OSS instructions too, BTW.

recipes/pack.sh
rafie: It's for packing the recipe, so one can simply deploy it on a host and enable the installation.

recipes/write_behind/oracle/rs/start-gear
rafie: It's a Redis Enterprise path, relevant to a runtime environment, not a build one.

recipes/write_behind/README.md
rafie: You don't need it, but when you need to install it on Redis Enterprise, it specifies how to do it.

recipes/write_behind/README.md
rafie: I'll add the OSS version.

MeirShpilraien:
@rafie still think we should remove the enterprise part as its not how enteprise users are going to install it.

recipes/write_behind/README.md
```diff
@@ -0,0 +1,20 @@
+# RedisGears Write-Behind Recipe
+
+## Deployment options
+
+* [With Redis OSS](oss/README.md)
+* [With Redis Enterprise Software and Oracle](oracle/rs/README.md)
+* [With Redis Enterprise Software and MySQL](mysql/rs/README.md)
```
MeirShpilraien: Still think we should drop the enterprise deployment doc for now cause this is not how its going to be in the end. We can save it as internal doc for internal usage ..

Issue #209, opened 2020-02-14
`make all` on macos catalina fails on


itamarhaber:
```
Packing dependencies...
./pack.sh: line 81: gawk: command not found
make: *** [artifacts/release/redisgears-dependencies.macosx-catalina-x64.99.99.99.tgz] Error 127
```

MeirShpilraien:
@rafie can we close this?

rafie:
#210 fxed the missing packages problem.
In CircleCI, we build for macOS mojave (10.14). I'll verify the catalina (10.15) build before we close this.

PR #210, opened 2020-02-16
Fixes for macOS build

rafie:


MeirShpilraien:
@rafie this should fix this? https://github.com/RedisGears/RedisGears/issues/209

PR #218, opened 2020-02-20
build: arm fixes and circleci/valgrind

rafie:


.circleci/config.yml
```diff
@@ -188,10 +218,16 @@ workflows:
           name: build-xenial
           platform: xenial
           <<: *platform-builds-flow-defs
+      - valgrind:
+          filters:
+            branches:
```
MeirShpilraien: Not sure we want it on each master commit, what do you think?

rafie: I guess you right. We can run it both daily and for version tags.

Issue #221, opened 2020-02-21
Make stream trimming optional.


shkediy:
When registering to a stream, after processing the stream item, gears trims the stream to size 0. It would be nice to have that as an option so that the stream can be kept in items need to be accessed later.

MeirShpilraien:
Fixed https://oss.redislabs.com/redisgears/master/readers.html#streamreader
Used trimStream option.

PR #222, opened 2020-02-22
see #221, added stream trimming as option to register function.

MeirShpilraien:


MeirShpilraien:
It is now possible to pass trimmStream variable as a parameter to register function
```
register('s2{06S}', trimmStream=False)
```

rafie:
trimmStream -> trimStream

PR #224, opened 2020-02-23
Added execution max idle parameter

MeirShpilraien:
Execution will be aborted if it was idle for more then the max idle value (by default 30 seconds).
If execution is aborted and messages about this exeuction is reached from other shards,
then it will be ignored, which means that eventually the execution will be timedout on all the shards.


MeirShpilraien:
- [ ] Add tests

- [ ] Run valgrind

MeirShpilraien:
Irrelevant, will be re-implemented after this https://github.com/RedisGears/RedisGears/pull/228 will be merge..

PR #228, opened 2020-02-27
Added thread pull support to run execution

MeirShpilraien:
1. We have a thread pull for execution run
2. It is possible to define a virtual worker.
   Execution on same worker will not run in parallel.
   Registrations use it to force execution order.
   Other execution run on their own worker, i.e, no execution order is prommised.

### DIFF
--- a/src/commands.c
+++ b/src/commands.c
@@ -173,7 +173,7 @@ Record* Command_AbortExecutionMap(ExecutionCtx* rctx, Record *data, void* arg){
         ExecutionPlan* gearsCtx = RedisGears_GetExecution(executionId);
 
         if(!gearsCtx){
-            RedisGears_SetError(rctx, "execution does not exists");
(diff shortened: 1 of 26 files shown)


MeirShpilraien:
- [ ] Run valgrind
- [ ] Add tests

src/commands.c
```diff
@@ -159,37 +163,46 @@ int Command_GetResultsBlocking(RedisModuleCtx *ctx, RedisModuleString **argv, in
...
+    while(true){
+        LockHandler_Acquire(ctx);
+        ExecutionPlan* gearsCtx = RedisGears_GetExecution(executionId);
 
-    if(!gearsCtx){
-        RedisModule_ReplyWithError(ctx, "Execution plan does not exits");
-        return REDISMODULE_OK;
-    }
+        if(!gearsCtx){
+            RedisGears_SetError(rctx, "execution does not exists");
```
rafie: exist

PR #232, opened 2020-03-01
Added deps/readies submodule

rafie:


MeirShpilraien:
@rafie do you expect the docker build to failed because of the submodule changes?

rafie:
Yes, it's a known issue in Dockerhub. If you add a submodule on a branch, it fails to check it out until you merge to masetr. After the merge, builds ok.

PR #251, opened 2020-03-10
[WIP] Coverage analysis support

rafie:


.circleci/config.yml
```diff
@@ -11,11 +11,17 @@ commands:
       test_params:
         type: string
         default: ""
-    steps: 
+    steps:
       - checkout
       - run:
           name: Submodule checkout
-          command: git submodule update --init --recursive
+          command: |
```
MeirShpilraien: @rafie can you explain what is this for?

.circleci/config.yml
```diff
@@ -188,6 +227,14 @@ workflows:
...
             tags:
               only: /.*/
+      - coverage:
+          requires:
+            - linux-debian
+          filters:
+            branches:
+              only: /.*/
+            tags:
+              only: /.*/
```
MeirShpilraien: How long is it taking, should we run it only on nightly?

build/scritps/valgrind.sh
```diff
@@ -0,0 +1,89 @@
+#!/bin/bash
```
MeirShpilraien: What is this script is used for?

.circleci/config.yml
rafie: I don't think it should take very long after I fix the mixup with the Valgrind build. It should be possible to run it per-commit.

MeirShpilraien: :+1:

.circleci/config.yml
rafie: This is an optimization for the dependency checkout. If we perform coverage analysis following the controlling build, we can save the dependencies in the build and retrieve them in the coverage phase. However, it looks that I might revert it if it turns out that the time gain doesn't worth the complexity.

MeirShpilraien: :+1:

MeirShpilraien:
@rafie Can this be merged?

build/scritps/valgrind.sh
rafie: This is used primarily for Callgrind invocation, and in general in any case you'd like to run valgrind without RLTest.

rafie:
@MeirShpilraien We're still missing some functionality here: 
- [ ] Prevent Valgrind involvement in coverage builds/runs
- [ ] Remove build artifacts caching from .circleci/config.yml

MeirShpilraien:
Use `make GCOV=1` to compile with coverage support and then `make coverage_report` to create the report.

PR #252, opened 2020-03-10
Deps installation

MeirShpilraien:


src/redisgears_python.c
```diff
@@ -65,145 +65,367 @@ static void* RedisGearsPy_PyCallbackDeserialize(FlatExecutionPlan* fep, Gears_Bu
...
+            ++c;
+        }
+        filesInDir = array_append(filesInDir, ' ');
     }
-    array_free(requirementsList);
+    filesInDir[array_len(filesInDir) - 1] = '\0';
+
+    int exitCode = ExecCommand(NULL, "/bin/bash -c \"source %s/bin/activate;cd %s;python -m pip install %s\"", venvDir, req->basePath, filesInDir);
+    array_free(filesInDir);
+    if(exitCode != 0){
```
rafie: return exitCode == 0;

src/redisgears_python.c
```diff
@@ -65,145 +65,367 @@ static void* RedisGearsPy_PyCallbackDeserialize(FlatExecutionPlan* fep, Gears_Bu
...
-        PythonSessionCtx* session = Gears_dictFetchValue(SessionsDict, id);
-        if(session){
-            PythonSessionCtx_FreeRequirementsList(requirementsList);
-            return PythonSessionCtx_ShellowCopy(session);
+static void PythonRequirementCtx_Free(PythonRequirementCtx* reqCtx){
+    if(--reqCtx->refCount){
+        return;
+    }
+
+    ExecCommand(NULL, "rm -rf %s", reqCtx->basePath);
```
rafie: reqCtx->basePath == "/" scares me.

src/redisgears_python.c
```diff
@@ -65,145 +65,367 @@ static void* RedisGearsPy_PyCallbackDeserialize(FlatExecutionPlan* fep, Gears_Bu
...
+
+            if(!PythonRequirementCtx_InstallRequirement(req)){
+                PythonRequirementCtx_Free(req);
+                PythonSessionCtx_Free(session);
+                session = NULL;
+                break;
+            }
+        }
+
+        session->requirements = array_append(session->requirements , req);
```
rafie: session can be NULL here.

MeirShpilraien: How can it be NULL?

src/redisgears_python.c
MeirShpilraien: What do you suggest?

src/redisgears_python.c
rafie: You're right, it seemed to me it was outside the for.

src/redisgears_python.c
rafie: From looking at the construction of basePath, it looks like we should check that it doesn't match /+. Maybe do it in a function and call it in creation site and rm -rf site just to be sure we don't rack the system.

PR #256, opened 2020-03-12
deps files will be read only on serialize deserialize

MeirShpilraien:


src/execution_plan.c
```diff
@@ -462,6 +446,35 @@ const char* FlatExecutionPlan_Serialize(FlatExecutionPlan* fep, size_t *len, cha
     return fep->serializedFep->buff;
 }
 
+int FlatExecutionPlan_Serialize(Gears_BufferWriter* bw, FlatExecutionPlan* fep, char** err){
+    // we serialize the PD of a fep each time cause it might be very big (contains file
```
rafie: Nowhere in the code is there an explanation of what a PD (Python Dependency?) really is. Maybe we should drop a comment in the header.

MeirShpilraien: Thing is that the private data could be anything but I agree I will add a more detailed explanation in the h file.

MeirShpilraien: Done

PR #263, opened 2020-03-23
Api fixes

MeirShpilraien:
Introduce the following api changes:

1. KeysOnlyReader is implemented as python reader and support the following arguments:
```
	def CreateKeysOnlyReader(pattern='*', count=1000, exactMatch=False, patternGenerator=None):
    	'''
    	Create a KeysOnlyReader callback as a python reader
    	pattern          - keys pattern to return
    	count            - count parameter given to scan command (ignored if isPattern is false)
    	exactMatch       - boolean indicating if exact match is require, i.e, do not use scan and just return the
                       give pattern
    	patternGenerator - a callbacks to generate different pattern on each shard. If given, the callback
                       will run on each shard and the return tuple (pattern, isPattern) will be used.
                       If this argument is given the pattern and isPattern arguments are ignored.
    	'''
```
2. KeysReader format was changed to (notice there is no backword issues here): {'key':..., 'value':..., 'type':..., 'event':...}
   * key - the key name (string)
   * value - the key value (type is depends on the key type)
   * type - the key type (string, hash, set, ...)
   * event - the event that trigger the exection (SET, HSET, ..), None on batch executions

   In addition the KeysReader introduce the following new/changed capabilities:
   * on 'run':
      * readValue - if False only the 'key' and the 'event' values will be provided (default True)
      * noScan - if True read the exect pattern given without performing scan operation (default False)
   * on 'register':
      * readValue - same as read value on 'run'
      * regex argument was depricated and changed to prefix

3. StreamReader format was changed to (notice there is a backword compatability): {'key':..., 'msg_id':..., 'value':{'k':'v',...}}
   * key - the stream key (string)
   * msg_id - the current message id (string)
   * value - the message value (dictionary)

   In addition the StreamReader introduce the following new/changed capabilities:
   * on 'run':
      * fronId - the id from which to start read messages from the stream (default 0-0)
   * on 'register':
      * regex argument was depricated and changed to prefix

4. the 'Log' function was changed to 'Log(msg, level='notice')'.
   If old argument format is used a depricated message is added to the log file

5. Pytest was added to cover all the new arguments

Still left to do:
- [x] Add tests for KeysOnlyReader
- [x] Document all the changes
- [x] Delete the KeysOnlyReader from Keys_reader.c
- [x] Change msg_id to just id

@itamarhaber FYI

### DIFF
--- a/docs/functions.md
+++ b/docs/functions.md
@@ -203,35 +203,21 @@ The **Register** action registers a function as an event handler. The function i
 
 **Python API**
 ```python
-class GearsBuilder.register(regex='*', mode='async', batch=1, duration=0,
-  eventTypes=None, keyTypes=None, onRegistered=None, onFailedPolicy="continue",
-  onFailedRetryInterval=1, trigger=None)
+class GearsBuilder.register(convertToStr=True, collect=True, mode='async', onRegistered=None)
 ```
 
 _Arguments_
 
-* _regex_: An optional argument that's passed to the reader as its _defaultArg_. It means the following:
-    * A key prefix for the [KeysReader](readers.md#keysreader) and [KeysOnlyReader](readers.md#keysonlyreaders) readers
-    * A key name or prefix for the [StreamReader](readers.md#streamreader) reader
+* _convertToStr_: when `True` adds a [map](operations.md#map) operation to the flow's end that stringifies records
+* _collect_: when `True` adds a [collect](operations.md#collect) operation to flow's end
+
 * _mode_: the execution mode of the triggered function. Can be one of:
     * **'async'**: execution will be asynchronous across the entire cluster
     * **'async_local'**: execution will be asynchronous and restricted to the handling shard
     * **'sync'**: execution will be synchronous and local
-* _batch_: the batch size that triggers execution for the [StreamReader](readers.md#streamreader) reader
-* _duration_: the interval between executions for the [StreamReader](readers.md#streamreader) reader (takes precedence over the _batch_ argument)
-* _eventTypes_: A whitelist of event types that trigger for the [KeysReader](readers.md#keysreader) or [KeysOnlyReader](readers.md#keysonlyreaders) readers. The list may contain one or more:
-    * Any Redis or module command
-    * Any [Redis event](https://redis.io/topics/notifications)
-* _keyTypes_: A whitelist of key types that trigger execution for the [KeysReader](readers.md#keysreader) or [KeysOnlyReader](readers.md#keysonlyreaders) readers. The list may contain one or more from the following:
-    * Redis core types: 'string', 'hash', 'list', 'set', 'zset' or 'stream'
-    * For module data types: 'module'
 * _onRegistered_: A function [callback](operations.md#callback) that's called on each shard upon function registration. It is a good place to initialize non-serializable objects such as network connections.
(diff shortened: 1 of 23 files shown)


src/GearsBuilder.py
```diff
@@ -15,6 +15,39 @@
...
+def CreateKeysOnlyReader(pattern='*', count=1000, exactMatch=False, patternGenerator=None):
+    '''
+    Create a KeysOnlyReader callback as a python reader
+    pattern          - keys pattern to return
+    count            - count parameter given to scan command (ignored if isPattern is false)
+    exactMatch       - boolean indicating if exact match is require, i.e, do not use scan and just return the
+                       give pattern
+    patternGenerator - a callbacks to generate different pattern on each shard. If given, the callback
+                       will run on each shard and the return tuple (pattern, isPattern) will be used.
+                       If this argument is given the pattern and isPattern arguments are ignored.
```
rafie: isPattern => exactMatch?

src/readers/keys_reader.c
```diff
@@ -124,78 +125,81 @@ static KeysReaderRegisterData* KeysReaderRegisterData_GetShallowCopy(KeysReaderR
...
+    RedisGears_BWWriteLong(bw, krctx->noScan);
+}
+
+static void RG_KeysReaderCtxDeserialize(FlatExecutionPlan* fep, void* ctx, Gears_BufferReader* br){
+    KeysReaderCtx* krctx = (KeysReaderCtx*)ctx;
+    krctx->match = RG_STRDUP(RedisGears_BRReadString(br));
+    if(RedisGears_BRReadLong(br)){
+        krctx->event = RG_STRDUP(RedisGears_BRReadString(br));
+        // having an even means this execution was triggered by a registration and the fact
+        // that we deserialize it means that the event did not accured on this shard,
```
rafie: accured => occurred

MeirShpilraien:
@rafie @itamarhaber can this be merged?

itamarhaber:
I'd like to also suggest:
[ ] Decapitalize `Log()` to `log()`
[ ] Ditto `ConfigGet` to `configGet()`

docs/functions.md
```diff
@@ -205,7 +205,7 @@ The **Register** action registers a function as an event handler. The function i
 ```python
 class GearsBuilder.register(regex='*', mode='async', batch=1, duration=0,
   eventTypes=None, keyTypes=None, onRegistered=None, onFailedPolicy="continue",
-  onFailedRetryInterval=1)
+  onFailedRetryInterval=1, trigger=None)
```
MeirShpilraien: @itamarhaber I am not sure about this session anymore, basically, the register gets a kargs now and each reader extracts what it needs.

MeirShpilraien:
@itamarhaber done

MeirShpilraien:
@rafie @itamarhaber let me know if I can merge

docs/functions.md
itamarhaber: I see.

So, are there any args that apply to all readers here?

itamarhaber:
lgtm :)

docs/functions.md
MeirShpilraien: Only the `onRegistered=None`, `mode`, `regex (which was changed to prefix)`
I will fix it

MeirShpilraien: Know what, its actually just the `onRegistered=None` because of command reader

PR #264, opened 2020-03-24
Extend time before failing due to lack of output

rafie:


MeirShpilraien:
@rafie why are you changing this change? I do not think we should reach so much time without output ..

rafie:
@MeirShpilraien It's probably only a macOS thing, and the built-in timeout (20m) is a magic number. If it's really stuck, it will fail over the 30m limit as well.

MeirShpilraien:
Yes but then we will have to wait 30 minutes. Regarding the failure yesterday, as I wrote on slack, it was really stuck and there is a bug there ..

rafie:
It may be a bug but the test does not fail consistently. I'll try to reproduce it on a non-CircleCI mac machine.

MeirShpilraien:
@rafie can we close this one?

PR #265, opened 2020-03-24
Minor fix for expected result

quintonparker:
Rick Sanchez is 70!


itamarhaber:
@meir this also needs to be cherry-picked to 0.9 :/

PR #266, opened 2020-03-24
Minor fix to expected result

quintonparker:
Rick is 70 and Morty is 14 therefore = 84


MeirShpilraien:
@itamarhaber this also needs to be cherry-picked to 0.9 right?

itamarhaber:
@MeirShpilraien totally.

PR #278, opened 2020-04-05
Fixes a false deprecation warning

itamarhaber:
When calling gearsConfigGet the warning gets printed.

Also, updates docs API snippets so these use the non-deprecated config methods and adds an example for automatically expiring keys.

### DIFF
--- a/docs/snippets/examples/automatic-expire.py
+++ b/docs/snippets/examples/automatic-expire.py
@@ -1,3 +1,3 @@
 gb = GB()
 gb.foreach(lambda x: execute('EXPIRE', x['key'], 3600))
-gb.register('*', readValue=False)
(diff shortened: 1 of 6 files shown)


docs/snippets/examples/automatic-expire.py
```diff
@@ -0,0 +1,3 @@
+gb = GB()
+gb.foreach(lambda x: execute('EXPIRE', x['key'], 3600))
+gb.register('*', readValue=False)
```
MeirShpilraien, line 3: mode='sync'?

itamarhaber: I guess it is correct-er to do that. Fixed.

PR #284, opened 2020-04-14
Add --requirements for requirements.txt

gkorland:


recipes/gears.py
```diff
@@ -41,6 +45,12 @@ def PP(res):
     print('Cannot connect to Redis. Aborting.')
     exit(1)
 
+if args.requirements is not None:
+    args.extra_args.append('REQUIREMENTS')
+    with open(args.requirements, 'r') as f:
+        requirements = [(el.strip()) for el in f.readlines()] 
+        args.extra_args = [*args.extra_args, *requirements]
```
MeirShpilraien, line 52: I think you can use +=, but its also fine ..

PR #300, opened 2020-05-01
Update subinterpreters.md

itamarhaber:


docs/subinterpreters.md
```diff
@@ -1,7 +1,7 @@
...
+One way to address this problem is by restarting the interpreter before each execution. However, an interpreter restart is a time-costly operation that renders this approach less-than-desirable. Also, two executions may be parallel, either in two different threads or even in the same thread (while the execution of the Python code itself is always non-parallelized, the rest of the execution's lifecycle may be). When more than one execution is running, restarting the interpreter is no longer an option. To address the issue correctly RedisGears plans to make use of [Python's Sub-Interpreters](https://docs.python.org/3/c-api/init.html#sub-interpreter-support).
```
MeirShpilraien, line 4: Maybe we should mentioned that currently we solve it with new global dictionary for each pyexecute (because we found out that there are libraries which are not compatible with subinyerpreters). Its not perfect but gives a good isolation and we are planning to return subinterpreters in the future?

itamarhaber: So repeat some of the stuff from https://github.com/RedisGears/RedisGears/blob/master/docs/runtime.md#python-interpreter - makes sense. The current is simply not true so we need to resolve it :)

MeirShpilraien: Yes I agree, we should call this page "runs isolation" or something and explain how its done today and what are the future plans. If you want I can take it.

mkdocs.yml
```diff
@@ -37,7 +37,7 @@ nav:
   - 'Examples': 'examples.md'
   - 'Clients': 'clients.md'
   - 'Design':
-    - 'Sub-Interpreters': 'subinterpreters.md'
+    - 'Isolation Technics': 'isolation_technics.md'
```
itamarhaber, line 40: @MeirShpilraien also needs to be renamed in runtime.md

MeirShpilraien: :+1:

PR #302, opened 2020-05-03
Fixes macOS build

rafie:


system-setup.py
```diff
@@ -108,6 +107,7 @@ def common_last(self):
         self.pip_install("--no-cache-dir git+https://github.com/Grokzen/redis-py-cluster.git@master")
         self.pip_install("--no-cache-dir git+https://github.com/RedisLabsModules/RLTest.git@master")
         self.pip_install("--no-cache-dir git+https://github.com/RedisLabs/RAMP@master")
+        self.pip_install("jinja2")
```
MeirShpilraien, line 110: What is this lib?

rafie: It's Python templating lib, needed to process ramp.yml via the readies xtx utility.

Makefile
```diff
@@ -371,7 +388,7 @@ test: __sep
 ifeq ($(DEBUG),1)
 	$(SHOW)set -e; cd pytest; VALGRIND=1 ./run_tests.sh
 else ifneq ($(TEST),)
-	@set -e; cd pytest; PYDEBUG=1 RLTest --test $(TEST) $(RLTEST_GDB) -s --module $(abspath $(TARGET))
+	@set -e; cd pytest; PYDEBUG=1 RLTest --test $(TEST) $(TEST_ARGS) $(RLTEST_GDB) -s --module $(abspath $(TARGET))
```
MeirShpilraien, line 391: What is this change?

Makefile
```diff
@@ -153,10 +158,20 @@ endif
 
 ifneq ($(OS),macosx)
 EMBEDDED_LIBS += -lutil
+else
+
```
MeirShpilraien, line 162: Empty else?

Makefile
```diff
@@ -153,10 +158,20 @@ endif
...
 endif
 
 endif # WITHPYTHON
 
+ifeq ($(SHOW_LD_LIBS),1)
+LD_FLAGS += -Wl,-t
```
MeirShpilraien, line 168: What is this feature?

rafie: Diagnostic feature. Prints all binary artifacts that go into the link.

Makefile
rafie: Enables one to specify extra args to `RLTest` (like when you want to run a specific test with `TEST` but want to also specify cluster type).

Makefile
```diff
@@ -127,6 +128,10 @@ endif
 
 ifeq ($(OS),macosx)
 LD_FLAGS += -undefined dynamic_lookup
+LD_FLAGS += \
+	-framework CoreFoundation \
+	$(GETTEXT_PREFIX)/lib/libintl.a \
```
rafie, line 133: Note that GETTEXT_PREFIX is set within `paella/mk/macosx.def` to be either `/usr/local` (for our custom build) or to the result of `brew --prefix gettext` (if we get the package with `brew`).

MeirShpilraien: Its only macos related right?

rafie: Yes, we don't currently install `libintl.a` (via `gettext`) for Linux. The reason we install it for macOS is that having the wrong `gettext` around (like the one you may get with `brew`) may break the build and there's no simple way of telling `cpython` to ignore `gettext` altogether.

.circleci/config.yml
```diff
@@ -227,12 +230,26 @@ workflows:
...
+#      - macos:
+#          name: macos-diag
 #          filters:
 #            branches:
-#              ignore: /.*/
+#              only: /.*/
 #            tags:
-#              only: /^v[0-9].*/
+#              only: /.*/
+#      - valgrind:
```
MeirShpilraien, line 240: Do we need this?

rafie: This is for running valgrind/macos from a PR (like, un-commenting for testing, then re-commenting before merge).

PR #308, opened 2020-05-05
fixes to work with RedisAI

MeirShpilraien:
@DvirDukhan FYI


MeirShpilraien:
> Regarding `GearsCheckSupportedVestion()`, don't you think failing for a mismatch in the patch version is a little too aggressive?

@rafie I do not think so we supported 5.0.7 only before so patch version does make a difference sometimes..

PR #316, opened 2020-05-10
system-setup: Ubuntu 20 fixes

rafie:


system-setup.py
```diff
@@ -35,7 +35,11 @@ def debian_compat(self):
         self.install("zip unzip")
 
         # pip cannot build gevent on ARM
-        self.install("python-psutil python-gevent")
+        self.install("python-psutil")
+        if self.os == 'ubuntu' and int(self.ver.split('.')) < 20:
```
MeirShpilraien, line 39: self.ver.split('.')[0]?

PR #319, opened 2020-05-11
Improved RAMP tool diagnostics

rafie:


MeirShpilraien:
@rafie any idea why docker failed?

rafie:
@MeirShpilraien In dockerhub it fails to build a ramp file, but doesn't print the redis output as expected, so I suspect it doesn't use the correct ramp version (in a way I find hard to explain). Currently building the docker locally. Will also re-run on dockerhub.

rafie:
Rebuild on dockerhub is good. Local docker build is also good.

PR #325, opened 2020-05-21
Fix cicling reference that cause logical memory leak

MeirShpilraien:


MeirShpilraien:
@rafie Any idea why docker fails?

rafie:
@MeirShpilraien Dockerhub logs still not available, will build locally and see.

rafie:
@MeirShpilraien Local docker build is good and dockerhub also succeeded.

PR #326, opened 2020-05-21
Requirement mgmt refactoring

MeirShpilraien:
Changes:
1. Added version to StepArgType (for future plane to get the python out of gear as a plugin)
2. Change requirement installation as follow:
	On get requirment such that there is a requirement that are not installed
	1. Download the wheels on background threa
	2. Create an execution to distribute and install the requirement on all the shards.
	   This task also send the requirement to slaves and replicate to aof.
        3. On done start run the execution code.
3. Introduce RG.PYIMPORTREQ, RG.PYEXPORTREQ to import and export registrations (for offline installation)
4. Introduce RG.PYDUMPREQS to dump all the installed requirements
5. Flat execution plane no longer serialized with their requirements to reduce network overhead.
   The assumption is that requirements exists on all the shards.
6. All the requirements are save to the rdb aux fields, the registrations serialize their requirements
   name and the pointer is resolved on load time. to ensure the requirements will deserialize before the
   registration we use the before keys rdb on requirements and the after key rdb for registrations.


MeirShpilraien:
- [ ] Add tests that verifies that flat execution plane that failed because requirement missing is recovered and not crashes the server.

docs/commands.md
```diff
@@ -18,6 +18,7 @@ The following sections describe the supported commands.
 | [`RG.INFOCLUSTER`](#rginfocluster) | Returns cluster information |
 | [`RG.PYEXECUTE`](#rgpyexecute) | Executes a Python function |
 | [`RG.PYSTATS`](#rgpystats) | Returns memory usage statistics |
+| [`RG.PYDUMPREQS`](#rgpystats) | Returns all the requirements (with information about each requirement) |
```
itamarhaber, line 21:
```suggestion
| [`RG.PYDUMPREQS`](#rgpystats) | Returns detailed information about requirements |
```

docs/commands.md
```diff
@@ -498,6 +499,41 @@ redis> RG.PYSTATS
...
+
+```
+RG.PYDUMPREQS
+```
+
+_Return_
+
+An array that consists of alternating key name and value entries as follows:
+
+* **GearReqVersion**: an internal version of the requirement given by redis gears (do not confuse between this and the requirement python version)
```
itamarhaber, line 515:
```suggestion
* **GearReqVersion**: an internally-assigned version of the requirement (note: this isn't the package's version)
```

docs/commands.md
```diff
@@ -498,6 +499,41 @@ redis> RG.PYSTATS
...
+* **Name**: the name of the requirement as it was given to the REQUIREMENTS argument on [RG.PYEXECUTE](commands.md#rgpyexecute) command
```
itamarhaber, line 516:
```suggestion
* **Name**: the name of the requirement as it was given to the `REQUIREMENTS` argument of the  [RG.PYEXECUTE](commands.md#rgpyexecute) command
```

docs/commands.md
```diff
@@ -498,6 +499,41 @@ redis> RG.PYSTATS
...
+* **IsDownloaded**: true if the requirement wheels was successfully download otherwise false
```
itamarhaber, line 517:
```suggestion
* **IsDownloaded**: `#!python True` if the requirement's wheels were successfully downloaded, `#!python False` otherwise
```

docs/commands.md
```diff
@@ -498,6 +499,41 @@ redis> RG.PYSTATS
...
+* **IsInstalled**: true if the requirement wheels was successfully installed otherwise false
```
itamarhaber, line 518:
```suggestion
* **IsInstalled**: `#!python True` if the requirement's wheels were successfully installed, `#!python False` otherwise
```

docs/commands.md
MeirShpilraien: Its not python True, just the text 'true', maybe I should change it to yes/no?

docs/commands.md
MeirShpilraien: Same comment as above

docs/commands.md
itamarhaber: Oh, right.

For RESP2 I'd stick with yes/no or 1/0. RESP3 has a Boolean type so that's a no-brainer.

docs/commands.md
itamarhaber: Agreed.

src/redisgears_python.c
```diff
@@ -146,10 +172,19 @@ static bool PythonRequirementCtx_DownloadRequirement(PythonRequirementCtx* req){
     }
     closedir(dr);
 
-    return true;
+    req->isDownloaded = true;
+
+done:
+    pthread_mutex_unlock(&req->installationLock);
```
rafie, line 178: should we also unlock before we assert (on the opendir)?

MeirShpilraien: Assert is not recoverable so why unlock?

docs/configuration.md
```diff
@@ -223,3 +223,18 @@ _Default Value_
...
+## PythonInstallReqMaxIdleTime
+The **PythonInstallReqMaxIdleTime** configuration option controls the maximal amount of idle time (in milliseconds) before python requirement installation is aborted. Idle time means no progress is made by the installation. The main reason for idle time is the same as for **ExecutionMaxIdleTime**.
+
+_Expected Value_
+
+Any integer greater than 0
+
+_Default Value_
+
+"30 seconds"
```
itamarhaber, line 236:
```suggestion
30000
```

docs/configuration.md
```diff
@@ -223,3 +223,18 @@ _Default Value_
 _Runtime Configurability_
 
 Supported
+
+## PythonInstallReqMaxIdleTime
+The **PythonInstallReqMaxIdleTime** configuration option controls the maximal amount of idle time (in milliseconds) before python requirement installation is aborted. Idle time means no progress is made by the installation. The main reason for idle time is the same as for **ExecutionMaxIdleTime**.
```
itamarhaber, line 228:
```suggestion
The **PythonInstallReqMaxIdleTime** configuration option controls the maximal amount of idle time (in milliseconds) before Python's requirements installation is aborted. Idle time means that the installation makes no progress. The main reason for idle time is the same as for **ExecutionMaxIdleTime**.
```

PR #330, opened 2020-05-24
do not crash on wrong reply status

MeirShpilraien:


MeirShpilraien:
- [ ] Add test to verify

docs/configuration.md
```diff
@@ -238,3 +238,18 @@ _Default Value_
 _Runtime Configurability_
 
 Supported
+
+## SendMsgRetries
+The **SendMsgRetries** configuration option controls the maximal amount of retries RedisGears will try to send a message before it gives up. Sent failure is considered either an error reply or a disconnection before getting ack on the message. 0 value indicates RedisGears to never give up.
```
itamarhaber, line 243:
```suggestion
The **SendMsgRetries** configuration option controls the maximum number of retries for sending a message between RedisGears' shards. When a message is sent and the shard disconnects before acknowledging it, or when it returns an error, the message will be resent until this threshold is met. Setting the value to 0 means unlimited retries.
```

PR #332, opened 2020-05-25
Reduce size of dokcer image & package sha testing

rafie:


Dockerfile
```diff
@@ -51,20 +54,19 @@ ARG PACK
...
-COPY --from=builder /build/artifacts/ /var/opt/redislabs/artifacts/
-
-RUN set -xe ;\
-	d=/var/opt/redislabs/modules/rg/ ;\
-	mkdir -p $d ;\
-	tar --warning=no-timestamp -xzf /var/opt/redislabs/artifacts/release/redisgears-dependencies.*.tgz -C $d
-
-RUN if [ "$PACK" != "1" ]; then rm -rf /var/opt/redislabs/artifacts; fi
+COPY --from=builder --chown=redis:redis /build/redisgears.so $REDIS_MODULES/
+COPY --from=builder --chown=redis:redis /build/bin/linux-x64-release/python3_* /var/opt/redislabs/modules/rg/python3/
```
MeirShpilraien, line 60: Shouldn't that be just /var/opt/redislabs/modules/rg/ ?

rafie: The whole thing is basically working against Dockerfile API. It does not allow you to do a parametric `COPY` operation.
If you specify just `/var/opt/redislabs/modules/rg/` it puts the content of `python3_*` into `.../rg/` without the `python3_*` dir.

MeirShpilraien: What I mean is that it should copy it to `/var/opt/redislabs/modules/rg/`

rafie: The workaround it to copy `python3_VERSION` from the builder's `bin/linux-x64-release` into `/var/opt/redislabs/modules/rg/python3`, and then to create a `python3_VERSION` symbolic link at `/var/opt/redislabs/modules/rg` to point to `python3`.
This way gears gets the dependency where it expect it and we don't waste image storage space (you cannot mv something without it requiring extra storage unless you squash - but dockerhub doesn't support it).

PR #345, opened 2020-06-01
Session str fix

MeirShpilraien:


src/common.c
```diff
@@ -130,15 +130,20 @@ int rg_asprintf(char **__ptr, const char *__restrict __fmt, ...) {
 char* ArrToStr(void** arr, size_t len, char*(*toStr)(void*)) {
     char* res = array_new(char, 100);
```
rafie, line 131: I'm a little bit worried about this 100.

PR #346, opened 2020-06-01
patched script run and execut command

DvirDukhan:
added bytes and bytearray argument to execute command. script returns list of tensors

### DIFF
--- a/pytest/test_basics.py
+++ b/pytest/test_basics.py
@@ -117,6 +117,20 @@ class testBasic:
         self.env.assertEqual(sum([a for a in range(100)]), int(res[0]))
         self.env.cmd('rg.dropexecution', id)
 
+def testBytes(env):
+    conn = getConnectionByEnv(env)
+    conn.set("x", 1)
+    conn.execute_command('rg.pyexecute', 'GB().repartition(lambda x: "y").foreach(lambda x: execute("set", "y", bytes([1,2,3]))).run("x")')
+    env.assertTrue(conn.exists("y"))
+
+
+def testBytearray(env):
+    conn = getConnectionByEnv(env)
+    conn.set("x", 1)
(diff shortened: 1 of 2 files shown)


pytest/test_basics.py
```diff
@@ -117,6 +117,20 @@ def testBasicAccumulate(self):
...
+    conn.execute_command('rg.pyexecute', 'GB().repartition(lambda x: "y").foreach(lambda x: execute("set", "y", bytes([1,2,3]))).run("x")')
+    env.assertTrue(conn.exists("y"))
+
+
+def testBytearray(env):
+    conn = getConnectionByEnv(env)
+    conn.set("x", 1)
+    conn.execute_command('rg.pyexecute', 'GB().repartition(lambda x: "y").foreach(lambda x: execute("set", "y", bytearray([1,2,3]))).run("x")')
+    res = conn.get("y")
+    env.assertEqual(bytearray([1,2,3]), res)
```
MeirShpilraien, line 132: conn.exists("y") ?

src/redisgears_python.c
```diff
@@ -1758,19 +1758,28 @@ static PyObject* executeCommand(PyObject *cls, PyObject *args){
...
+        if(PyByteArray_Check(argument)) {
+            argLen = PyByteArray_Size(argument);
+            argumentCStr = PyByteArray_AsString(argument);
+        } else if(PyBytes_Check(argument)) {
+            argLen = PyBytes_Size(argument);
+            argumentCStr = PyBytes_AsString(argument);
+        } else {
+            PyObject* argumentStr = PyObject_Str(argument);
+            argumentCStr = PyUnicode_AsUTF8AndSize(argumentStr, &argLen);
+            Py_DECREF(argumentStr);
```
MeirShpilraien, line 1775: The decrease ref must come after you use the underline pointer. Otherwise there is a potential issue that the pointer might be freed on decreasing ref count..

PR #350, opened 2020-06-04
ai API modifications

DvirDukhan:
added AI API:
1. Tensor get/set from keyspace - without the need to use rm_call
2. variadic input for scripts

Tested on demo

### DIFF
--- a/src/redisgears_python.c
+++ b/src/redisgears_python.c
@@ -2093,6 +2093,90 @@ error:
     return NULL;
 }
 
+static PyObject* setTensorInKey(PyObject *cls, PyObject *args) {
+    verifyRedisAILoaded();
+    // Input validation: 2 arguments. args[0]: string. args[1]: tensor
+    if(PyTuple_Size(args) != 2){
+        PyErr_SetString(GearsError, "Wrong number of arguments given to setTensorInKey");
+        return NULL;
+    }
+    PyObject* keyName = PyTuple_GetItem(args, 0);
+    if(!PyUnicode_Check(keyName)){
+        PyErr_SetString(GearsError, "key name argument must be a string");
+        return NULL;
+    }
+    PyTensor* pyt = (PyTensor*)PyTuple_GetItem(args, 1);
+    if(!PyObject_IsInstance((PyObject*)pyt, (PyObject*)&PyTensorType)){
+        PyErr_SetString(GearsError, "Given argument is not of type PyTensor");
+        return NULL;
+    }
+
+    RedisModuleType *ai_tensorType = RedisAI_TensorRedisType();
+    size_t len;
+    const char* keyNameCStr = PyUnicode_AsUTF8AndSize(keyName, &len);
+
+    RedisModuleCtx* rctx = RedisModule_GetThreadSafeContext(NULL);
+    RedisModuleString *keyNameStr = RedisModule_CreateString(rctx, keyNameCStr, len);
+    LockHandler_Acquire(rctx);
+    RedisModuleKey *key = RedisModule_OpenKey(rctx,keyNameStr, REDISMODULE_WRITE);
+    // Set a shllow copy of the tensor.
+    RedisModule_ModuleTypeSetValue(key, ai_tensorType, RedisAI_TensorGetShallowCopy(pyt->t));
+    LockHandler_Release(rctx);
+    RedisModule_FreeString(rctx, keyNameStr);
+    RedisModule_FreeThreadSafeContext(rctx);
+    return PyLong_FromLong(1);
+}
+
+static PyObject* getTensorFromKey(PyObject *cls, PyObject *args) {
+     verifyRedisAILoaded();
+    // Input validation: 1 argument. args[0]: string.
+    if(PyTuple_Size(args) != 1){
+        PyErr_SetString(GearsError, "Wrong number of arguments given to setTensorInKey");
+        return NULL;
+    }
+    PyObject* keyName = PyTuple_GetItem(args, 0);
+    if(!PyUnicode_Check(keyName)){
+        PyErr_SetString(GearsError, "key name argument must be a string");
+        return NULL;
+    }
+
+    RedisModuleType *ai_tensorType = RedisAI_TensorRedisType();
+    size_t len;
+    const char* keyNameCStr = PyUnicode_AsUTF8AndSize(keyName, &len);
+
+    RedisModuleCtx* rctx = RedisModule_GetThreadSafeContext(NULL);
+    RedisModuleString *keyNameStr = RedisModule_CreateString(rctx, keyNameCStr, len);
+    LockHandler_Acquire(rctx);
+    RedisModuleKey *key = RedisModule_OpenKey(rctx,keyNameStr, REDISMODULE_READ);
+    // Empty key.
+    if (RedisModule_KeyType(key) == REDISMODULE_KEYTYPE_EMPTY) {
+        RedisModule_CloseKey(key);
+        LockHandler_Release(rctx);
+        RedisModule_FreeString(rctx, keyNameStr);
+        RedisModule_FreeThreadSafeContext(rctx);
+        PyErr_SetString(GearsError, "ERR tensor key is empty");
+        return NULL;
+    }
+    // No tensor key.
+    if (RedisModule_ModuleTypeGetType(key) != ai_tensorType) {
+        RedisModule_CloseKey(key);
+        LockHandler_Release(rctx);
+        RedisModule_FreeString(rctx, keyNameStr);
+        RedisModule_FreeThreadSafeContext(rctx);
+        PyErr_SetString(GearsError, "ERR key is not a tensor");
+        return NULL;
+    }
+    // Read tensor and free locks.
+    RAI_Tensor *tensor = RedisModule_ModuleTypeGetValue(key);
+    LockHandler_Release(rctx);
+    RedisModule_FreeString(rctx, keyNameStr);
+    RedisModule_FreeThreadSafeContext(rctx);
+
+    PyTensor* pyt = PyObject_New(PyTensor, &PyTensorType);
+    pyt->t = RedisAI_TensorGetShallowCopy(tensor);
+    return (PyObject*)pyt;
+}
+
 typedef struct PyGraphRunner{
    PyObject_HEAD
    RAI_ModelRunCtx* g;
@@ -2365,7 +2449,59 @@ static PyObject* scriptRunnerAddInput(PyObject *cls, PyObject *args){
         PyErr_SetString(GearsError, "Given argument is not of type PyTensor");
         return NULL;
     }
-    RedisAI_ScriptRunCtxAddInput(pys->s, pyt->t);
+    RAI_Error* err;
+    RedisAI_InitError(&err);
+    RedisAI_ScriptRunCtxAddInput(pys->s, pyt->t, err);
+    if (RedisAI_GetErrorCode(err) != RedisAI_ErrorCode_OK) {
+        PyErr_SetString(GearsError, RedisAI_GetError(err));
+        RedisAI_FreeError(err);
+        return NULL;
+    }
+    RedisAI_FreeError(err);
+    return PyLong_FromLong(1);
+}
+
+static PyObject* scriptRunnerAddInputList(PyObject *cls, PyObject *args){
+    verifyRedisAILoaded();
+    if(PyTuple_Size(args) != 2){
+        PyErr_SetString(GearsError, "Wrong number of arguments given to scriptRunnerAddInput");
+        return NULL;
+    }
+    PyTorchScriptRunner* pys = (PyTorchScriptRunner*)PyTuple_GetItem(args, 0);
+    if(!PyObject_IsInstance((PyObject*)pys, (PyObject*)&PyTorchScriptRunnerType)){
+        PyErr_SetString(GearsError, "Given argument is not of type PyTorchScriptRunner");
+        return NULL;
+    }
+
+    PyObject* py_tensorList = (PyListObject*)PyTuple_GetItem(args, 1);
+    if(!PyObject_IsInstance(py_tensorList, (PyObject*)&PyList_Type)){
+        PyErr_SetString(GearsError, "Given argument is not of type PyList");
+        return NULL;
+    }
+
+    size_t len = PyList_Size(py_tensorList);
+    RAI_Tensor** tensorList = array_new(RAI_Tensor*, len);
+    for(size_t i=0; i < len; i++){
+        PyTensor* pyt = (PyTensor*)PyList_GetItem(py_tensorList, i);
+        if(!PyObject_IsInstance((PyObject*)pyt, (PyObject*)&PyTensorType)){
+            PyErr_SetString(GearsError, "Given argument is not of type PyTensor");
+            array_free(tensorList);
+            return NULL;
+        }
+        tensorList = array_append(tensorList, pyt->t);
+    }
+
+    RAI_Error* err;
+    RedisAI_InitError(&err);
+    RedisAI_ScriptRunCtxAddInputList(pys->s, tensorList, array_len(tensorList), err);
+    if (RedisAI_GetErrorCode(err) != RedisAI_ErrorCode_OK) {
+        PyErr_SetString(GearsError, RedisAI_GetError(err));
+        RedisAI_FreeError(err);
+        array_free(tensorList);
+        return NULL;
+    }
+    array_free(tensorList);
+    RedisAI_FreeError(err);
     return PyLong_FromLong(1);
 }
 
@@ -2625,12 +2761,15 @@ PyMethodDef EmbRedisGearsMethods[] = {
 PyMethodDef EmbRedisAIMethods[] = {
     {"createTensorFromValues", createTensorFromValues, METH_VARARGS, "creating a tensor object from values"},
     {"createTensorFromBlob", createTensorFromBlob, METH_VARARGS, "creating a tensor object from blob"},
+    {"setTensorInKey", setTensorInKey, METH_VARARGS, "set a tensor in keyspace"},
+    {"getTensorFromKey", getTensorFromKey, METH_VARARGS, "get a tensor from keyspace"},
     {"createModelRunner", createModelRunner, METH_VARARGS, "open TF graph by key name"},
     {"modelRunnerAddInput", modelRunnerAddInput, METH_VARARGS, "add input to graph runner"},
     {"modelRunnerAddOutput", modelRunnerAddOutput, METH_VARARGS, "add output to graph runner"},
     {"modelRunnerRun", modelRunnerRun, METH_VARARGS, "run graph runner"},
     {"createScriptRunner", createScriptRunner, METH_VARARGS, "open a torch script by key name"},
     {"scriptRunnerAddInput", scriptRunnerAddInput, METH_VARARGS, "add input to torch script runner"},
+    {"scriptRunnerAddInputList", scriptRunnerAddInputList, METH_VARARGS, "add a list of tensor as input to torch script runner"},
     {"scriptRunnerAddOutput", scriptRunnerAddOutput, METH_VARARGS, "add output to torch script runner"},
     {"scriptRunnerRun", scriptRunnerRun, METH_VARARGS, "run torch script runner"},
     {"tensorToFlatList", tensorToFlatList, METH_VARARGS, "turning tensor into flat list"},
--- a/src/redisgears_python.c
+++ b/src/redisgears_python.c
@@ -2164,24 +2164,15 @@ static PyObject* msetTensorsInKeyspace(PyObject *cls, PyObject *args) {
         return NULL;
     }
 
+    PyObject* obj = NULL;
     bool error = false;
-    bool heap_allocation = false;
     bool is_cluster = Cluster_IsClusterMode();
     size_t len = PyDict_Size(py_tensorDict);
     size_t stack_buffer_size = 1000;
-    RAI_Tensor* tensor_stack_buffer[stack_buffer_size];
-    const char* key_name_buffer[stack_buffer_size];
-    RAI_Tensor** tensorList;
-    const char** keyNameList;
-    if(len <= stack_buffer_size) {
-        tensorList = tensor_stack_buffer;
-        keyNameList = key_name_buffer;
-    }
-    else {
-        tensorList = RedisModule_Alloc(len * sizeof(RAI_Tensor*));
-        keyNameList = RedisModule_Alloc(len * sizeof(const char*));
-        heap_allocation = true;
-    }
+    array_new_on_stack(RedisModuleKey*, stack_buffer_size, keys);
+    array_new_on_stack(RAI_Tensor*, stack_buffer_size, tensorList);
+    array_new_on_stack(const char*, stack_buffer_size, keyNameList);
+
     Py_ssize_t i = 0;
     PyObject *key, *value;
     // Validate each key-value pair for <string, tensor> type.
@@ -2196,7 +2187,7 @@ static PyObject* msetTensorsInKeyspace(PyObject *cls, PyObject *args) {
             error = true;
             break;          
         }
-        tensorList[i-1] = ((PyTensor*)value)->t;
+        tensorList = array_append(tensorList, ((PyTensor*)value)->t);
         size_t keylen;
         const char* keyNameCStr = PyUnicode_AsUTF8AndSize(key, &keylen);
         if(is_cluster){
@@ -2207,28 +2198,14 @@ static PyObject* msetTensorsInKeyspace(PyObject *cls, PyObject *args) {
                 break;
             }
         }
-        keyNameList[i-1] = keyNameCStr;
+        keyNameList = array_append(keyNameList, keyNameCStr);
     }
-    if(error){
-        if(heap_allocation){
-            RedisModule_Free(tensorList);
-            RedisModule_Free(keyNameList);
-        }
-        return NULL;
-    }
-    PyObject* obj = NULL;
-    RedisModuleKey* keys_stack_buffer[stack_buffer_size];
-    RedisModuleKey** keys;
-    if(heap_allocation){
-        keys = keys_stack_buffer;
-    }
-    else {
-        keys = RedisModule_Alloc(len * sizeof(RedisModuleKey*));
-    }
+    if(error) goto clean_up;
+   
     RedisModuleType *ai_tensorType = RedisAI_TensorRedisType();
     RedisModuleCtx* rctx = RedisModule_GetThreadSafeContext(NULL);
     LockHandler_Acquire(rctx);
-    for(i = 0; i < len; i++) {
+    for(size_t i = 0; i < len; i++) {
         const char *keyNameCStr = keyNameList[i];
         RedisModuleString *keyNameStr = RedisModule_CreateString(rctx, keyNameCStr, strlen(keyNameCStr));
         RedisModuleKey *key = RedisModule_OpenKey(rctx,keyNameStr, REDISMODULE_WRITE);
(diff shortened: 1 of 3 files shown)


src/redisgears_python.c
```diff
@@ -2093,6 +2093,90 @@ static PyObject* createTensorFromValues(PyObject *cls, PyObject *args){
...
+    }
+
+    RedisModuleType *ai_tensorType = RedisAI_TensorRedisType();
+    size_t len;
+    const char* keyNameCStr = PyUnicode_AsUTF8AndSize(keyName, &len);
+
+    RedisModuleCtx* rctx = RedisModule_GetThreadSafeContext(NULL);
+    RedisModuleString *keyNameStr = RedisModule_CreateString(rctx, keyNameCStr, len);
+    LockHandler_Acquire(rctx);
+    RedisModuleKey *key = RedisModule_OpenKey(rctx,keyNameStr, REDISMODULE_WRITE);
```
MeirShpilraien, line 2121: Isn't there an open key that gets char*

src/redisgears_python.c
```diff
@@ -2093,6 +2093,90 @@ static PyObject* createTensorFromValues(PyObject *cls, PyObject *args){
...
+    // Set a shllow copy of the tensor.
+    RedisModule_ModuleTypeSetValue(key, ai_tensorType, RedisAI_TensorGetShallowCopy(pyt->t));
```
MeirShpilraien, line 2123: What if the key is not empty?

src/redisgears_python.c
```diff
@@ -2093,6 +2093,90 @@ static PyObject* createTensorFromValues(PyObject *cls, PyObject *args){
...
+    LockHandler_Release(rctx);
+    RedisModule_FreeString(rctx, keyNameStr);
+    RedisModule_FreeThreadSafeContext(rctx);
+    return PyLong_FromLong(1);
```
MeirShpilraien, line 2127: Let's return PyNone, make sure to increase it refcount before returning it.

src/redisgears_python.c
```diff
@@ -2093,6 +2093,90 @@ static PyObject* createTensorFromValues(PyObject *cls, PyObject *args){
...
+        PyErr_SetString(GearsError, "key name argument must be a string");
+        return NULL;
+    }
+
+    RedisModuleType *ai_tensorType = RedisAI_TensorRedisType();
+    size_t len;
+    const char* keyNameCStr = PyUnicode_AsUTF8AndSize(keyName, &len);
+
+    RedisModuleCtx* rctx = RedisModule_GetThreadSafeContext(NULL);
+    RedisModuleString *keyNameStr = RedisModule_CreateString(rctx, keyNameCStr, len);
```
MeirShpilraien, line 2148: Save about the char* comment, I am not sure there is such api though ..

src/redisgears_python.c
```diff
@@ -2093,6 +2093,90 @@ static PyObject* createTensorFromValues(PyObject *cls, PyObject *args){
...
+        RedisModule_CloseKey(key);
+        LockHandler_Release(rctx);
+        RedisModule_FreeString(rctx, keyNameStr);
+        RedisModule_FreeThreadSafeContext(rctx);
+        PyErr_SetString(GearsError, "ERR key is not a tensor");
+        return NULL;
+    }
+    // Read tensor and free locks.
+    RAI_Tensor *tensor = RedisModule_ModuleTypeGetValue(key);
+    LockHandler_Release(rctx);
```
MeirShpilraien, line 2171: You should only free the lock after you get the ShallowCopy, otherwise someone might delete the key (and free the tensor) before you get the ShallowCopy.

src/redisgears_python.c
```diff
@@ -2365,7 +2449,59 @@ static PyObject* scriptRunnerAddInput(PyObject *cls, PyObject *args){
...
+    RAI_Error* err;
+    RedisAI_InitError(&err);
+    RedisAI_ScriptRunCtxAddInput(pys->s, pyt->t, err);
+    if (RedisAI_GetErrorCode(err) != RedisAI_ErrorCode_OK) {
+        PyErr_SetString(GearsError, RedisAI_GetError(err));
+        RedisAI_FreeError(err);
+        return NULL;
+    }
+    RedisAI_FreeError(err);
+    return PyLong_FromLong(1);
```
MeirShpilraien, line 2461: Lets return PyNone here as well

src/redisgears_python.c
```diff
@@ -2365,7 +2449,59 @@ static PyObject* scriptRunnerAddInput(PyObject *cls, PyObject *args){
...
+        return NULL;
+    }
+    PyTorchScriptRunner* pys = (PyTorchScriptRunner*)PyTuple_GetItem(args, 0);
+    if(!PyObject_IsInstance((PyObject*)pys, (PyObject*)&PyTorchScriptRunnerType)){
+        PyErr_SetString(GearsError, "Given argument is not of type PyTorchScriptRunner");
+        return NULL;
+    }
+
+    PyObject* py_tensorList = (PyListObject*)PyTuple_GetItem(args, 1);
+    if(!PyObject_IsInstance(py_tensorList, (PyObject*)&PyList_Type)){
```
MeirShpilraien, line 2477: You should use PyList_Check

MeirShpilraien: Also, why not use iterable? why force it to a list?

src/redisgears_python.c
```diff
@@ -2365,7 +2449,59 @@ static PyObject* scriptRunnerAddInput(PyObject *cls, PyObject *args){
...
+    RedisAI_InitError(&err);
+    RedisAI_ScriptRunCtxAddInputList(pys->s, tensorList, array_len(tensorList), err);
+    if (RedisAI_GetErrorCode(err) != RedisAI_ErrorCode_OK) {
+        PyErr_SetString(GearsError, RedisAI_GetError(err));
+        RedisAI_FreeError(err);
+        array_free(tensorList);
+        return NULL;
+    }
+    array_free(tensorList);
+    RedisAI_FreeError(err);
```
MeirShpilraien, line 2504: I know it was like this before but lets also return PyNone

src/redisgears_python.c
```diff
@@ -2365,7 +2449,59 @@ static PyObject* scriptRunnerAddInput(PyObject *cls, PyObject *args){
...
+            PyErr_SetString(GearsError, "Given argument is not of type PyTensor");
+            array_free(tensorList);
+            return NULL;
+        }
+        tensorList = array_append(tensorList, pyt->t);
+    }
+
+    RAI_Error* err;
+    RedisAI_InitError(&err);
+    RedisAI_ScriptRunCtxAddInputList(pys->s, tensorList, array_len(tensorList), err);
```
MeirShpilraien, line 2496: We need to consider who is taking and giving ownership here, does RedisAI get ShallowCopy of all the tensors in the list?

src/redisgears_python.c
DvirDukhan: `void *REDISMODULE_API_FUNC(RedisModule_OpenKey)(RedisModuleCtx *ctx, RedisModuleString *keyname, int mode);
`

src/redisgears_python.c
DvirDukhan: I don't think we should return an error. I think we should allow re-write keys

MeirShpilraien: I disagree, at least it should be a flag. And even if you override you must release the old value otherwise you are leaking...

DvirDukhan: ok, I see what you mean.

src/redisgears_python.c
DvirDukhan: redisAI generates a shallow copy

MeirShpilraien: Well rethinking about it now, the ref count is not increased atomically so theoretically we have an issue here and we need to take the GIL before call this function (same for scriptRunnerAddInput).

src/redisgears_python.c
```diff
@@ -2093,6 +2094,309 @@ static PyObject* createTensorFromValues(PyObject *cls, PyObject *args){
...
 }
 
+static PyObject* setTensorInKey(PyObject *cls, PyObject *args) {
+    verifyRedisAILoaded();
+    // The function returns Py_None in any case, prepare it.
+    Py_INCREF(Py_None);
+    // Input validation: 2 arguments. args[0]: string. args[1]: tensor
+    if(PyTuple_Size(args) != 2){
+        PyErr_SetString(GearsError, "Wrong number of arguments given to setTensorInKey");
+        return Py_None;
```
MeirShpilraien, line 2104: If there is an error you need to return NULL

src/redisgears_python.c
```diff
@@ -2093,6 +2094,309 @@ static PyObject* createTensorFromValues(PyObject *cls, PyObject *args){
...
+
+    RedisModuleType *ai_tensorType = RedisAI_TensorRedisType();
+    size_t len;
+    const char* keyNameCStr = PyUnicode_AsUTF8AndSize(keyName, &len);
+
+    if(Cluster_IsClusterMode()){
+        const char* keyNodeId = Cluster_GetNodeIdByKey(keyNameCStr);
+        if(!Cluster_IsMyId(keyNodeId)){
+            PyErr_SetString(GearsError, "Given key is not in the current shard");
+            Py_INCREF(Py_None);
```
MeirShpilraien, line 2127: You increased Py_None twice ... and regardless you need to return NULL

src/redisgears_python.c
MeirShpilraien: Still not see where you free the old value if exists

src/redisgears_python.c
```diff
@@ -2093,6 +2094,309 @@ static PyObject* createTensorFromValues(PyObject *cls, PyObject *args){
...
+    // Set a shllow copy of the tensor.
+    RedisModule_ModuleTypeSetValue(key, ai_tensorType, RedisAI_TensorGetShallowCopy(pyt->t));
+    RedisModule_CloseKey(key);
+    LockHandler_Release(rctx);
+    RedisModule_FreeString(rctx, keyNameStr);
+    RedisModule_FreeThreadSafeContext(rctx);
+    return Py_None;
+}
+
+static PyObject* msetTensorsInKeyspace(PyObject *cls, PyObject *args) {
```
MeirShpilraien, line 2155: Why you need also mset and setTensorInKey. Isn't mset cover set?

MeirShpilraien: I would think about this all function design again. Lots of things can be allocated on the stack instead of array_new, I suggest to have a goto error and clean the memory there...

src/redisgears_python.c
```diff
@@ -2093,6 +2094,309 @@ static PyObject* createTensorFromValues(PyObject *cls, PyObject *args){
...
+    PyObject* py_tensorDict = (PyListObject*)PyTuple_GetItem(args, 0);
+    if(!PyDict_Check(py_tensorDict)){
+        PyErr_SetString(GearsError, "Given argument is not of type PyDict");
+        return Py_None;
+    }
+    bool error = false;
+    bool is_cluster = Cluster_IsClusterMode();
+    size_t len = PyDict_Size(py_tensorDict);
+    RAI_Tensor** tensorList = array_new(RAI_Tensor*, len);
+    const char** keyNameList = array_new(const char*, len);
```
MeirShpilraien, line 2174: Can be allocated on stack ..

src/redisgears_python.c
```diff
@@ -2093,6 +2094,309 @@ static PyObject* createTensorFromValues(PyObject *cls, PyObject *args){
...
+            }
+        }
+        keyNameList = array_append(keyNameList, keyNameCStr);
+    }
+    if(error) {
+        array_free(keyNameList);
+        array_free(tensorList);
+        return Py_None;
+    }
+    RedisModuleKey** keys = array_new(RedisModuleKey*, len);
```
MeirShpilraien, line 2207: Can be allocated on stack ..

src/redisgears_python.c
```diff
@@ -2093,6 +2094,309 @@ static PyObject* createTensorFromValues(PyObject *cls, PyObject *args){
...
+
+}
+
+static PyObject* getTensorFromKey(PyObject *cls, PyObject *args) {
+     verifyRedisAILoaded();
+    // Input validation: 1 argument. args[0]: string.
+    if(PyTuple_Size(args) != 1){
+        PyErr_SetString(GearsError, "Wrong number of arguments given to getTensorFromKey");
+        Py_INCREF(Py_None);
+        return Py_None;
```
MeirShpilraien, line 2250: Return NULL

src/redisgears_python.c
```diff
@@ -2093,6 +2094,309 @@ static PyObject* createTensorFromValues(PyObject *cls, PyObject *args){
...
+        LockHandler_Release(rctx);
+        RedisModule_FreeString(rctx, keyNameStr);
+        RedisModule_FreeThreadSafeContext(rctx);
+        PyErr_SetString(GearsError, "ERR tensor key is empty");
+        Py_INCREF(Py_None);
+        return Py_None;
+    }
+    // No tensor key.
+    if (RedisModule_ModuleTypeGetType(key) != ai_tensorType) {
+        RedisModule_CloseKey(key);
```
MeirShpilraien, line 2286: goto error

src/redisgears_python.c
```diff
@@ -2093,6 +2094,309 @@ static PyObject* createTensorFromValues(PyObject *cls, PyObject *args){
...
+    RedisModule_FreeString(rctx, keyNameStr);
+    RedisModule_FreeThreadSafeContext(rctx);
+
+    return (PyObject*)pyt;
+}
+
+static PyObject* mgetTensorsFromKeyspace(PyObject *cls, PyObject *args) {
+     verifyRedisAILoaded();
+    // Input validation: 1 argument. args[0]: List<String>.
+    if(PyTuple_Size(args) != 1){
```
MeirShpilraien, line 2308: I also believe lots of this code can be shared with mset ...

src/redisgears_python.c
```diff
@@ -2351,22 +2655,79 @@ static PyObject* createScriptRunner(PyObject *cls, PyObject *args){
 
 static PyObject* scriptRunnerAddInput(PyObject *cls, PyObject *args){
     verifyRedisAILoaded();
+    // The function returns Py_None in any case, prepare it.
+    Py_INCREF(Py_None);
     if(PyTuple_Size(args) != 2){
         PyErr_SetString(GearsError, "Wrong number of arguments given to scriptRunnerAddInput");
-        return NULL;
+        return Py_None;
```
MeirShpilraien, line 2662: This was correct actually :)

src/redisgears_python.c
```diff
@@ -2351,22 +2655,79 @@ static PyObject* createScriptRunner(PyObject *cls, PyObject *args){
...
+    }
+
+    RAI_Tensor** tensorList = array_new(RAI_Tensor*, 10);
+    PyTensor* pyt;
+    while(pyt=PyIter_Next(py_tensors_iter)){
+        if(!PyObject_IsInstance((PyObject*)pyt, (PyObject*)&PyTensorType)){
+            PyErr_SetString(GearsError, "Given argument is not of type PyTensor");
+            array_free(tensorList);
+        return Py_None;
```
MeirShpilraien, line 2714: Decrease py_tensors_iter ref count

src/redisgears_python.c
```diff
@@ -2351,22 +2655,79 @@ static PyObject* createScriptRunner(PyObject *cls, PyObject *args){
...
+        }
+        tensorList = array_append(tensorList, pyt->t);
+    }
+
```
MeirShpilraien, line 2718: Decrease py_tensors_iter ref count

src/redisgears_python.c
DvirDukhan: I don't mind removing the `set`, I thought about `mset` after I had `set` implemented, so I didn't want to delete

also, all lists are now stack allocated arrays

src/redisgears_python.c
```diff
@@ -2093,6 +2094,334 @@ static PyObject* createTensorFromValues(PyObject *cls, PyObject *args){
...
+    PyObject* obj = NULL;
+    RedisModuleCtx* rctx = RedisModule_GetThreadSafeContext(NULL);
+    RedisModuleString *keyNameStr = RedisModule_CreateString(rctx, keyNameCStr, len);
+    LockHandler_Acquire(rctx);
+    RedisModuleKey *key = RedisModule_OpenKey(rctx,keyNameStr, REDISMODULE_WRITE);
+    // Check if the key is empty or 
+    bool empty = RedisModule_KeyType(key) == REDISMODULE_KEYTYPE_EMPTY;
+    if(!empty && (RedisModule_ModuleTypeGetType(key) != ai_tensorType)) {
+        PyErr_SetString(GearsError, "ERR Key is already set with non tensor values.");
+        goto clean_up;
```
MeirShpilraien, line 2138: :+1:

src/redisgears_python.c
```diff
@@ -2093,6 +2094,334 @@ static PyObject* createTensorFromValues(PyObject *cls, PyObject *args){
...
+}
+
+static PyObject* msetTensorsInKeyspace(PyObject *cls, PyObject *args) {
+    verifyRedisAILoaded();
+    if(PyTuple_Size(args) != 1){
+        PyErr_SetString(GearsError, "Wrong number of arguments given to msetTensorInKeyspace");
+        return NULL;
+    }
+    // Input should be a Dictionary<string, tensor>.
+    PyObject* py_tensorDict = (PyListObject*)PyTuple_GetItem(args, 0);
```
MeirShpilraien, line 2161: You do not need the cast

src/redisgears_python.c
```diff
@@ -2093,6 +2094,291 @@ static PyObject* createTensorFromValues(PyObject *cls, PyObject *args){
...
+            PyErr_SetString(GearsError, "Given key is not in the current shard");
+            return NULL;
+        }
+    }
+
+    PyObject* obj = NULL;
+    RedisModuleCtx* rctx = RedisModule_GetThreadSafeContext(NULL);
+    RedisModuleString *keyNameStr = RedisModule_CreateString(rctx, keyNameCStr, len);
+    LockHandler_Acquire(rctx);
+    RedisModuleKey *key = RedisModule_OpenKey(rctx,keyNameStr, REDISMODULE_WRITE);
```
MeirShpilraien, line 2133: I think the return key might be NULL no?

src/redisgears_python.c
```diff
@@ -2093,6 +2094,291 @@ static PyObject* createTensorFromValues(PyObject *cls, PyObject *args){
...
+    while (PyDict_Next(py_tensorDict, &i, &key, &value)) {
+        if(!PyUnicode_Check(key)){
+            PyErr_SetString(GearsError, "key name argument must be a string");
+            error = true;
+            break;
+        }
+        if(!PyObject_IsInstance(value, (PyObject*)&PyTensorType)){
+            PyErr_SetString(GearsError, "Given argument is not of type PyTensor");
+            error = true;
+            break;
```
MeirShpilraien, line 2188: Why not goto clean_up and then you do not need the error variable?

src/redisgears_python.c
```diff
@@ -2093,6 +2094,291 @@ static PyObject* createTensorFromValues(PyObject *cls, PyObject *args){
...
+    if(!PyDict_Check(py_tensorDict)){
+        PyErr_SetString(GearsError, "Given argument is not of type PyDict");
+        return NULL;
+    }
+
+    PyObject* obj = NULL;
+    bool error = false;
+    bool is_cluster = Cluster_IsClusterMode();
+    size_t len = PyDict_Size(py_tensorDict);
+    size_t stack_buffer_size = 1000;
```
MeirShpilraien, line 2171: Define? And I would put it to 100, 1000 is a lot on the stack ..

src/redisgears_python.c
```diff
@@ -2093,6 +2094,291 @@ static PyObject* createTensorFromValues(PyObject *cls, PyObject *args){
...
+    array_new_on_stack(RedisModuleKey*, stack_buffer_size, keys);
```
MeirShpilraien, line 2172: I think ArrayDefine would be a better name here, what do you think?

src/redisgears_python.c
```diff
@@ -2093,6 +2094,291 @@ static PyObject* createTensorFromValues(PyObject *cls, PyObject *args){
...
+    array_new_on_stack(RAI_Tensor*, stack_buffer_size, tensorList);
+    array_new_on_stack(const char*, stack_buffer_size, keyNameList);
+
+    Py_ssize_t i = 0;
+    PyObject *key, *value;
+    // Validate each key-value pair for <string, tensor> type.
+    while (PyDict_Next(py_tensorDict, &i, &key, &value)) {
```
MeirShpilraien, line 2179: What the difference between this and get iterator ? Could we still run on this dictionary after?

src/redisgears_python.c
```diff
@@ -2093,6 +2094,291 @@ static PyObject* createTensorFromValues(PyObject *cls, PyObject *args){
...
+        return NULL;
+        }
+    }
+    PyObject* obj = NULL;
+    RedisModuleType *ai_tensorType = RedisAI_TensorRedisType();
+    RedisModuleCtx* rctx = RedisModule_GetThreadSafeContext(NULL);
+    RedisModuleString *keyNameStr = RedisModule_CreateString(rctx, keyNameCStr, len);
+    LockHandler_Acquire(rctx);
+    RedisModuleKey *key = RedisModule_OpenKey(rctx,keyNameStr, REDISMODULE_READ);
+    bool error = false;
```
MeirShpilraien, line 2268: You do not use this

src/redisgears_python.c
```diff
@@ -2093,6 +2094,291 @@ static PyObject* createTensorFromValues(PyObject *cls, PyObject *args){
...
+    size_t stack_buffer_size = 1000;
+    array_new_on_stack(const char*, stack_buffer_size, keyNameList);
+    array_new_on_stack(RedisModuleKey*, stack_buffer_size, keys);
+    bool is_cluster = Cluster_IsClusterMode();
+    PyObject* keyName = NULL;
+    while((keyName = PyIter_Next(keys_iter)) != NULL) {
+        // Check for string.
+        if(!PyUnicode_Check(keyName)) {
+            PyErr_SetString(GearsError, "key name argument must be a string");
+            error = true;
```
MeirShpilraien, line 2317: Again why not call the goto here?

src/redisgears_python.c
```diff
@@ -2093,6 +2094,291 @@ static PyObject* createTensorFromValues(PyObject *cls, PyObject *args){
...
+    }
+    obj = tensorsList;
+
+clean_up:
+
+    array_free_ex(keys, RedisModule_CloseKey(*(RedisModuleKey**)ptr));
+    LockHandler_Release(rctx);
+    RedisModule_FreeThreadSafeContext(rctx);
+    array_free(keyNameList);
+    array_free(keys);
```
MeirShpilraien, line 2378: Does the array_free_ex free the keys?

src/redisgears_python.c
```diff
@@ -2365,8 +2651,63 @@ static PyObject* scriptRunnerAddInput(PyObject *cls, PyObject *args){
...
+    }
+
+     PyObject* py_tensors_iter = PyObject_GetIter(PyTuple_GetItem(args, 1));
+    if(!py_tensors_iter){
+        PyErr_SetString(GearsError, "Tensors argument must be iterable");
+        return NULL;
+    }
+
+    PyObject* obj = NULL;
+    array_new_on_stack(RAI_Tensor*, 1000, tensorList);
```
MeirShpilraien, line 2687: Again I think 1000 is to much

src/redisgears_python.c
```diff
@@ -2365,8 +2651,63 @@ static PyObject* scriptRunnerAddInput(PyObject *cls, PyObject *args){
...
+    if (RedisAI_GetErrorCode(err) != RedisAI_ErrorCode_OK) {
+        PyErr_SetString(GearsError, RedisAI_GetError(err));
+        goto clean_up;
+    }
+    Py_INCREF(Py_None);
+    obj = Py_None;
+clean_up:
+    Py_DECREF(py_tensors_iter);
+    array_free(tensorList);
+    RedisAI_FreeError(err);
```
MeirShpilraien, line 2709: Error might not have been initialized here

src/utils/arr.h
```diff
@@ -73,13 +75,44 @@ static array_t array_new_sz(uint32_t elem_sz, uint32_t cap, uint32_t len) {
...
 /* Initialize an array for a given type T with a given length. The capacity allocated is identical
  * to the length
  *  */
 #define array_newlen(T, len) (array_new_sz(sizeof(T), len, len))
 
+static inline array_t persist(array_t arr){
+  array_hdr_t *hdr = array_hdr(arr);
+  if(!hdr->on_stack) return arr;
+  array_t tmp = array_new_sz(hdr->elem_sz, hdr->cap, hdr->len);
+  memcpy(tmp, arr, hdr->cap);
```
MeirShpilraien, line 101: shouldn't we just use adr->len ? also on like above? There is no need to allocate the entire capacity? what do you think?

src/utils/arr.h
```diff
@@ -73,13 +75,44 @@ static array_t array_new_sz(uint32_t elem_sz, uint32_t cap, uint32_t len) {
...
+  varName ## _struct.cap = (capacity);              \
+  varName ## _struct.elem_sz = sizeof(T);           \
+  T* varName = varName ## _struct.buf;
+
 /* Initialize an array for a given type T with a given length. The capacity allocated is identical
  * to the length
  *  */
 #define array_newlen(T, len) (array_new_sz(sizeof(T), len, len))
 
+static inline array_t persist(array_t arr){
```
MeirShpilraien, line 97: lets call it array_persist?

src/redisgears_python.c
DvirDukhan: you open with WRITE so no

src/redisgears_python.c
DvirDukhan: I think that the users should know that they explicitly allocating a dynamic array on the stack. There is no better notification to the user than this.

src/redisgears_python.c
DvirDukhan: I saw this example at `dictobject.c`. I believe this is a managed iteration over the data structure so. As far as I understand the code it depends on the value of `i` so It is reusable.

MeirShpilraien: Ohh right I missed the i, :+1:

src/redisgears_python.c
MeirShpilraien: So maybe stack_array_new?

MeirShpilraien: Though the new suggest that its a function and its not, that's why I thought we should call it define, how about stack_array_define?

src/redisgears_python.c
```diff
@@ -2093,6 +2096,277 @@ static PyObject* createTensorFromValues(PyObject *cls, PyObject *args){
...
+    for(size_t i = 0; i < len; i++) {
+        const char *keyNameCStr = keyNameList[i];
+        RedisModuleString *keyNameStr = RedisModule_CreateString(rctx, keyNameCStr, strlen(keyNameCStr));
+        RedisModuleKey *key = RedisModule_OpenKey(rctx,keyNameStr, REDISMODULE_WRITE);
+        // Check if the key is empty or 
+        bool empty = RedisModule_KeyType(key) == REDISMODULE_KEYTYPE_EMPTY;
+        if(!empty && (RedisModule_ModuleTypeGetType(key) != ai_tensorType)) {
+            RedisModule_FreeString(rctx, keyNameStr);
+            PyErr_SetString(GearsError, "ERR Key is already set with non tensor values.");
+            goto clean_up;
```
MeirShpilraien, line 2213: If you go to clean_up here do not free the rest of the keys in the keyNameList no? Isn't it better to use array_free_ex for keyNameList also?

src/redisgears_python.c
```diff
@@ -2093,6 +2096,277 @@ static PyObject* createTensorFromValues(PyObject *cls, PyObject *args){
...
+        if (RedisModule_KeyType(key) == REDISMODULE_KEYTYPE_EMPTY) {
+            RedisModule_FreeString(rctx, keyNameStr);
+            PyErr_SetString(GearsError, "ERR tensor key is empty");
+            goto clean_up;
+        }
+        // No tensor key.
+        if (RedisModule_ModuleTypeGetType(key) != ai_tensorType) {
+            RedisModule_FreeString(rctx, keyNameStr);
+            PyErr_SetString(GearsError, "ERR key is not a tensor");
+            goto clean_up;
```
MeirShpilraien, line 2343: Same here about the keyNameList

src/redisgears_python.c
DvirDukhan: keyNemeList contains `const char*` I don't need to free it

MeirShpilraien: right. nice catch ..

PR #360, opened 2020-06-11
added buffer abilities to tensor from blob

DvirDukhan:


src/redisgears_python.c
```diff
@@ -2007,12 +2007,20 @@ static PyObject* createTensorFromBlob(PyObject *cls, PyObject *args){
         // Blob is byte array.
         size = PyByteArray_Size(pyBlob);
         blob = PyByteArray_AsString(pyBlob);
```
MeirShpilraien, line 2009: So I guess you need the RedisAI_TensorSetData here as well right?

PR #373, opened 2020-06-25
registered execution plans

DvirDukhan:


src/execution_plan.c
```diff
@@ -358,6 +358,10 @@ static int FlatExecutionPlan_SerializeStep(FlatExecutionStep* step, Gears_Buffer
     return FlatExecutionPlan_SerializeStepArg(type, step->bStep.arg.stepArg, bw, err);
 }
 
+static inline void FlatExecutionPlan_SerializeID(FlatExecutionPlan* fep, Gears_BufferWriter* bw) {
+    RedisGears_BWWriteBuffer(&bw, fep->id, ID_LEN);
```
MeirShpilraien, line 362: bw not &bw

src/execution_plan.c
```diff
@@ -443,6 +447,16 @@ static const char* FlatExecutionPlan_SerializeInternal(FlatExecutionPlan* fep, s
 }
 
 int FlatExecutionPlan_Serialize(Gears_BufferWriter* bw, FlatExecutionPlan* fep, char** err){
+
```
MeirShpilraien, line 450: You need to put it in the FlatExecutionPlan_Distributed, this function is also used by rdb save and on rdb save we do want to save the flat execution ..

src/execution_plan.c
```diff
@@ -598,8 +617,16 @@ static int FlatExecutionPlan_DeserializeInternal(FlatExecutionPlan* ret, const c
 }
 
 FlatExecutionPlan* FlatExecutionPlan_Deserialize(Gears_BufferReader* br, char** err, int encver){
-    FlatExecutionPlan* ret = FlatExecutionPlan_New();
+    bool registered = RedisGears_BRReadLong(br);
```
MeirShpilraien, line 620: Should be done on ExecutionPlan_OnReceived

src/execution_plan.c
```diff
@@ -598,8 +617,16 @@ static int FlatExecutionPlan_DeserializeInternal(FlatExecutionPlan* ret, const c
...
+    if(registered) {
+            char *id = FlatExecutionPlan_DeserializeID(br);
+            FlatExecutionPlan* fep =  Gears_dictFetchValue(epData.registeredFepDict, id);
```
MeirShpilraien, line 623: FlatExecutionPlan_FindId

src/execution_plan.c
```diff
@@ -598,8 +617,16 @@ static int FlatExecutionPlan_DeserializeInternal(FlatExecutionPlan* ret, const c
...
+            assert(fep);
```
MeirShpilraien, line 624: No assert, we do not want to crash here, raise error

src/execution_plan.c
```diff
@@ -598,8 +617,16 @@ static int FlatExecutionPlan_DeserializeInternal(FlatExecutionPlan* ret, const c
...
+            RG_FREE(id);
```
MeirShpilraien, line 625: No need to free it, its part of the buffer ...

src/execution_plan.c
```diff
@@ -2342,6 +2369,7 @@ int FlatExecutionPlan_Register(FlatExecutionPlan* fep, ExecutionMode mode, void*
     RedisModule_Replicate(ctx, RG_INNER_REGISTER_COMMAND, "b", buff->buff, buff->size);
     RedisModule_FreeThreadSafeContext(ctx);
     Gears_BufferFree(buff);
+    FlatExecutionPlan_SetRegistered(fep);
```
MeirShpilraien, line 2372: Should be done on FlatExecutionPlan_RegisterInternal which called also on slaves (and all the shards in the cluster ..)

MeirShpilraien: Also why you need a set function for it and not just set it, if you chose a function make if static ..

src/execution_plan.h
```diff
@@ -260,6 +260,7 @@ typedef struct FlatExecutionPlan{
     FlatBasicStep onRegisteredStep;
     FlatBasicStep onUnpausedStep;
     long long executionMaxIdleTime;
+    bool registered;
```
MeirShpilraien, line 263: Lets make it execution flags, might be more of those in the future

DvirDukhan: ok
still keeping values as true / false

src/execution_plan.h
```diff
@@ -260,6 +261,7 @@ typedef struct FlatExecutionPlan{
     FlatBasicStep onRegisteredStep;
     FlatBasicStep onUnpausedStep;
     long long executionMaxIdleTime;
+    ExecutionFlags flags;
```
MeirShpilraien, line 264: ExecutionFlags are for executions, lets create a FlatExecutionFlags

src/redisgears_python.c
```diff
@@ -1983,7 +1983,7 @@ static PyObject* createTensorFromBlob(PyObject *cls, PyObject *args){
     bool buffered = false;
     Py_buffer view;
     size_t size;
-    const char* blob;
+    char* blob;
```
MeirShpilraien, line 1986: Why is this needed?

src/execution_plan.c
```diff
@@ -1892,16 +1909,25 @@ static void ExecutionPlan_UnregisterExecutionReceived(RedisModuleCtx *ctx, const
...
+    char* err = NULL;
     Gears_BufferReader br;
     Gears_BufferReaderInit(&br, &buff);
-    char* err = NULL;
-    FlatExecutionPlan* fep = FlatExecutionPlan_Deserialize(&br, &err, REDISGEARS_DATATYPE_VERSION);
+    bool registered = RedisGears_BRReadLong(&br);
+    FlatExecutionPlan* fep;
+    if(registered) {
+        char *id = FlatExecutionPlan_DeserializeID(&br); 
+        fep = FlatExecutionPlan_FindId(id);
```
MeirShpilraien, line 1926: If fep == NULL set err here so it will be printed to the log message

src/execution_plan.c
```diff
@@ -642,6 +650,15 @@ static void ExecutionPlan_Distribute(ExecutionPlan* ep){
...
+        if(EPIsFlagOn(ep->fep, EFRegistered)){
+        // Registered execution plan - serialize id and return.
+        RedisGears_BWWriteLong(&bw, 1);
+        FlatExecutionPlan_SerializeID(ep->fep, &bw);
+        return;
+    } else {
+        RedisGears_BWWriteLong(&bw, 0); // Non Registered execution plan.
+    }
+    
     int res = FlatExecutionPlan_Serialize(&bw, ep->fep, NULL);
```
MeirShpilraien, line 662: Notice that you serialize the FEP anyway

src/redisgears_python.c
DvirDukhan: compile warning

src/execution_plan.h
DvirDukhan: ok
I'm leaving `EPTurnOnFlag` or rename too?

src/execution_plan.c
DvirDukhan: Do I need to serialize? 
I thought that we just serializing ID

MeirShpilraien: Yes but this line happened anyway in you code:
FlatExecutionPlan_Serialize(...

MeirShpilraien: Ohh sorry, missed the return ..

MeirShpilraien: But this is an issue, if you return you do not serialize the reader data. This is why the tests fails ..

src/execution_plan.c
```diff
@@ -1507,17 +1524,13 @@ static void ExecutionPlan_RegisterForRun(ExecutionPlan* ep){
 
 void FlatExecutionPlan_AddToRegisterDict(FlatExecutionPlan* fep){
     Gears_dictAdd(epData.registeredFepDict, fep->id, fep);
-    // call the on registered callback if set
-    if(fep->onRegisteredStep.stepName){
```
MeirShpilraien, line 1511: Why you remove this code?

src/execution_plan.c
```diff
@@ -1529,6 +1542,12 @@ static int FlatExecutionPlan_RegisterInternal(FlatExecutionPlan* fep, RedisGears
     // the registeredFepDict holds a weak pointer to the fep struct. It does not increase
     // the refcount and will be remove when the fep will be unregistered
     FlatExecutionPlan_AddToRegisterDict(fep);
+    // call the on registered callback if set
```
MeirShpilraien, line 1545: Why you returned this code here?

src/execution_plan.c
```diff
@@ -1507,17 +1524,13 @@ static void ExecutionPlan_RegisterForRun(ExecutionPlan* ep){
 
 void FlatExecutionPlan_AddToRegisterDict(FlatExecutionPlan* fep){
     Gears_dictAdd(epData.registeredFepDict, fep->id, fep);
-    // call the on registered callback if set
-    if(fep->onRegisteredStep.stepName){
-        RedisGears_FlatExecutionOnRegisteredCallback onRegistered = FlatExecutionOnRegisteredsMgmt_Get(fep->onRegisteredStep.stepName);
-        RedisModule_Assert(onRegistered);
-        onRegistered(fep, fep->onRegisteredStep.arg.stepArg);
-    }
+    EPTurnOnFlag(fep, FEFRegistered);
```
MeirShpilraien, line 1527: This macro is for ExecutionPlan, We should create one from FlatExecutionPlan or change the name (remove the EP prefix)

src/execution_plan.c
DvirDukhan: bad rebase?

src/execution_plan.c
DvirDukhan: bad rebase?

src/execution_plan.c
DvirDukhan: added ones for FEP - copy paste for now, but if we need changes it is nice to have

docs/commands.md
```diff
@@ -152,6 +152,7 @@ An array with an entry per execution. Each entry is made of alternating key name
 
 * **executionId**: the [execution ID](functions.md#execution-id)
 * **status**: the [status](functions.md#execution-status)
+* **registered execution**: Indication that the given execution is a registered execution plan.
```
MeirShpilraien, line 155: An indication that the given execution was triggered by a registration?

@itamarhaber what do you think?

itamarhaber: 1. I don't like the field name having a space and being so long, would replace it with `registered`
2. (note the lowercase i and no ending period)
```suggestion
* **registered**: indicates whether this is a [registered execution](functions.html#registration)
```

MeirShpilraien:
@DvirDukhan did you run a full valgrind tests?

PR #374, opened 2020-06-28
Fix RedisAI to flat list function to handle long long encoding

MeirShpilraien:


MeirShpilraien:
Found here: https://forum.redislabs.com/t/svm-model-yields-different-outputs-on-redis-gear/629/6

PR #379, opened 2020-07-07
shardUniqueId when logfile is not present

rafie:


src/common.c
```diff
@@ -201,16 +202,29 @@ const char* GetShardUniqueId() {
...
         const char* logFileName = RedisModule_CallReplyStringPtr(uuidReply, &len);
-        const char* last = strrchr(logFileName, '/');
-        if(last){
-            len = len - (last - logFileName + 1);
-            logFileName = last + 1;
+        char uuid[64];
+        if (*logFileName == '\0' || *logFileName == '\n' || *logFileName == '\r' || *logFileName == ' ') {
+            RedisModule_Log(ctx, "notice", "No log file is available");
+            uuid_t binuuid;
+            uuid_generate_random(binuuid);
```
MeirShpilraien, line 212: @rafie uuid is problematic because it will be different each time you start the redis

src/common.c
```diff
@@ -201,16 +202,46 @@ const char* GetShardUniqueId() {
...
-            len = len - (last - logFileName + 1);
-            logFileName = last + 1;
+        char uuid[64];
+        bool found = true;
+        if (*logFileName == '\0' || *logFileName == '\n' || *logFileName == '\r' || *logFileName == ' ') {
+            //dbfilename
+            RedisModule_FreeCallReply(reply);
+            RedisModule_Log(ctx, "notice", "No log file is available");
+
+            reply = RedisModule_Call(ctx, "CONFIG", "cc", "GET", "logfile");
```
MeirShpilraien, line 215: Not follow, you try to check the log file again?

rafie:
Ok, so I added the log collection from the platform builds, and disabled my changes.
The problem does not reproduce - Redis responds with a non-empty log path.
I'm putting my changes back now.

PR #406, opened 2020-10-05
Merge master to 1.0 2

MeirShpilraien:


rafie (approved):
Checked vs master. Looks identical, apart from the right places (few docs and version.h).

PR #409, opened 2020-10-05
Clean alerts

gkorland:


src/GearsBuilder.py
```diff
@@ -1,10 +1,6 @@
 import redisgears
 import copy
-import redisgears as rg
 from redisgears import executeCommand as execute
-from redisgears import atomicCtx as atomic
```
MeirShpilraien, line 5: So you do not want our users to have atomic and hashtag functions?

recipes/write_behind/write_behind.py
```diff
@@ -3,7 +3,6 @@
 VERSION = '99.99.99'
 NAME = 'WriteBehind'
 
-import time
```
MeirShpilraien, line 6: This entire file need to be deleted anyway...

gkorland: Is that the effect? I was just removing unused imports

MeirShpilraien: Its not unused, its unused in this file, do not believe those static analyzers....

MeirShpilraien:
@gkorland can I close this?

MeirShpilraien:
@gkorland ^^

PR #421, opened 2020-10-23
Adds missing period / full stop.

simonprickett:


gkorland (approved):

MeirShpilraien:
@simonprickett we did a big update to the docs, I think this is no longer relevant, let me know if you agree (maybe I missed something) and if we can close.

simonprickett:
> @simonprickett we did a big update to the docs, I think this is no longer relevant, let me know if you agree (maybe I missed something) and if we can close.

Sounds good @MeirShpilraien, closing.

PR #426, opened 2020-10-26
Add the reliable-Keyspace

gkorland:


docs/snippets/examples/reliable-keyspace-notification.py
```diff
@@ -0,0 +1,3 @@
+GearsBuilder() \
+.foreach(lambda x: execute('XADD', "notifications-stream", '*', *sum([[k,v] for k,v in x.items()],[]))) \
+.register(prefix="person:*", eventTypes=['hset', 'hmset'])
```
MeirShpilraien, line 3: What about the other hset commands? del? rename?

MeirShpilraien (commented):
@gkorland why not creating a recipe that handles all the edge cases like rename/del maybe can handle more datatypes?

docs/snippets/examples/reliable-keyspace-notification.py
gkorland: it's just an example, similar to using `prefix="person:*"`

gkorland:
@MeirShpilraien I'll generate a full recipe too, it's just nice simple example

MeirShpilraien (approved):

PR #431, opened 2020-11-02
Better Redis-Enterprise cluster integration

MeirShpilraien:
1. The following is not allow if cluster is not yet initalized:
   * RG.PYEXECUTE
   * RG.DROPEXECUTION
   * RG.ABORTEXECUTION
   * RG.UNREGISTER
2. Import python requirement will not be allowed if cluster is not yet initalized.
3. RG.TRIGGER that triggered distributed execution will return error if cluster
   is not yet initalized.
4. Key space notification that trigger a distributed execution will failed.
5. A shard that did not yet recieve the cluster topology will send 'ERRCLUSTER'
   error, a shard that gets this error will send the cluster topology to the
   other shard so the cluster will get back to functional state faster
   and we will be able to continue running executions.
6. hashtag() function will use RedisModule_ShardingGetSlotRange RedisModule API
   (if available) to not be depends on ClusterSet command they might arrive later.

### DIFF
--- a/src/cluster.c
+++ b/src/cluster.c
@@ -14,6 +14,8 @@
 #include "config.h"
 #include <libevent.h>
 
+#define CLUSTER_SET_MY_ID_INDEX 6
+
 // forward declaration
 static void RG_HelloResponseArrived(struct redisAsyncContext* c, void* a, void* b);
 
@@ -37,6 +39,7 @@ typedef struct Node{
     NodeStatus status;
     struct event *reconnectEvent;
     struct event *resendHelloMessage;
+    bool sendClusterTopologyOnNextConnect;
 }Node;
 
 Gears_dict* nodesMsgIds;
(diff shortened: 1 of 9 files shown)


MeirShpilraien:
- [ ] cherrypick to 1.0

src/cluster.c
```diff
@@ -233,6 +249,12 @@ static void OnResponseArrived(struct redisAsyncContext* c, void* a, void* b){
         return;
     }
     Node* n = (Node*)b;
+    if(reply->type == REDIS_REPLY_ERROR && strncmp(reply->str, CLUSTER_ERROR, strlen(CLUSTER_ERROR)) == 0){
```
rafie, line 252: `strncmp(reply->str, CLUSTER_ERROR, strlen(CLUSTER_ERROR))` vs `strcmp(reply->str, CLUSTER_ERROR)`

MeirShpilraien: @rafie only the prefix need to be compared

rafie (commented):
LGTM. Minor comment.

rafie (approved):

PR #448, opened 2020-11-24
Fix missuse on async api that might cause crashes

MeirShpilraien:


MeirShpilraien:
- [ ] require retro review by @rafie

rafie (approved):

PR #457, opened 2020-12-08
Fix crash on stream reader when stream are dropped while processed.

MeirShpilraien:


rafie (approved):

MeirShpilraien:
Cherrypick to 1.0 : https://github.com/RedisGears/RedisGears/pull/458

PR #458, opened 2020-12-08
Fix crash on stream reader when stream are dropped while processed.

MeirShpilraien:
(cherry picked from commit f7523702c5e302c8453f4f1de8b9e627d3145938)

### DIFF
--- a/pytest/test_register.py
+++ b/pytest/test_register.py
@@ -987,8 +987,6 @@ GB("StreamReader").foreach(lambda x: time.sleep(2)).register(prefix='s')
     '''
     env.expect('rg.pyexecute', script).ok()
 
-    verifyRegistrationIntegrity(env)
-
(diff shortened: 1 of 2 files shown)


pytest/test_register.py
```diff
@@ -976,4 +976,23 @@ def testKeyReaderRegisterDontReadValues(env):
                     break
                 time.sleep(0.1)
     except Exception as e:
-        env.assertTrue(False, message='Failed waiting for list to popultae')
\ No newline at end of file
+        env.assertTrue(False, message='Failed waiting for list to popultae')
+
+def testDeleteStreamDurringRun(env):
```
rafie, line 981: typo: During

MeirShpilraien: Fixed

rafie (approved):

PR #461, opened 2020-12-19
Extract python to be gears plugin

MeirShpilraien:


plugins/python/redisgears_python.c
```diff
@@ -5544,8 +5546,8 @@ static int PythonConfig_GetStrVal(const char* val, const char* defaultVal, const
 #define ForeceDownloadDepsOnEnterpriseConfigName "ForeceDownloadDepsOnEnterprise"
 #define PythonAttemptTracebackConfigName "PythonAttemptTraceback"
 #define PythonInstallReqMaxIdleTimeConfigName "PythonInstallReqMaxIdleTime"
-#define DependenciesUrlConfigName "DependenciesUrl"
-#define DependenciesSha256ConfigName "DependenciesSha256"
+#define GearsPythonUrlConfigName "GearsPythonUrl"
+#define GearsPythonSha256ConfigName "GearsPythonSha256"
```
MeirShpilraien, line 5550: The name itself can not be changed cause it breaks backword compatability.

leakcheck.supp
```diff
@@ -1,3 +1,10 @@
+{
```
MeirShpilraien, line 1: Why is this needed?

plugins/python/redisgears_python.c
rafie: I'll revert that.

leakcheck.supp
rafie: It pops up in macOS. Just to clear noise. It's not essential, I can remove it.

MeirShpilraien: But maybe its not noise?

rafie: That's possible. I can reproduce it and we'll look into it.

rafie (approved):

PR #463, opened 2020-12-22
Enable async run of RedisAI model through python script

alonre24:


src/redisgears_python.c
```diff
@@ -3034,6 +3034,76 @@ static PyObject* modelRunnerRun(PyObject *cls, PyObject *args){
...
+
+	PyList_Append(l, tensorList);
+	PyList_Append(l, errs);
+
+	if (RedisAI_GetErrorCode(error) != RedisAI_ErrorCode_OK) {
+		const char *errStr = RedisAI_GetError(error);
+		PyObject* pyErr = PyUnicode_FromStringAndSize(errStr, strlen(errStr));
+		PyList_Append(errs, pyErr);
+		Py_DECREF(pyErr);
+		goto finish;
```
MeirShpilraien, line 3057: So if there are errors there is no way there will also be results? If so maybe we should expose `set_exception` (https://docs.python.org/3/library/asyncio-future.html) and expose it to C just like we expose `setFutureResultsFunction`

If there is a possibility for results and errors then this code ignores results when there is an error.

MeirShpilraien (commented):
Looks very good, small comment/suggestion.

src/redisgears_python.c
```diff
@@ -3034,6 +3035,69 @@ static PyObject* modelRunnerRun(PyObject *cls, PyObject *args){
...
+	if(!PyObject_IsInstance((PyObject*)pyg, (PyObject*)&PyGraphRunnerType)){
+		PyErr_SetString(GearsError, "Given argument is not of type PyGraphRunner");
+		return NULL;
+	}
+
+	PyObject* pArgs = PyTuple_New(0);
+	PyObject* future = PyObject_CallObject(createFutureFunction, pArgs);
+	Py_DECREF(pArgs);
+
+	RedisAI_ModelRunAsync(pyg->g, FinishAsyncModelRun, future);
```
MeirShpilraien, line 3096: is pyg->g refcounted? if so do we need to increase refcount or does RedisAI_ModelRunAsync will do it? If it's not refcounted then is it ok that it will be freed before the modelrun will finished?

src/redisgears_python.c
```diff
@@ -3025,15 +3040,83 @@ static PyObject* modelRunnerRun(PyObject *cls, PyObject *args){
     }
     RedisAI_FreeError(err);
     PyObject* tensorList = PyList_New(0);
-    for(size_t i = 0 ; i < RedisAI_ModelRunCtxNumOutputs(pyg->g) ; ++i){
+    for(size_t i = 0 ; i < RedisAI_ModelRunCtxNumOutputs(mctx) ; ++i){
         PyTensor* pyt = PyObject_New(PyTensor, &PyTensorType);
-        pyt->t = RedisAI_TensorGetShallowCopy(RedisAI_ModelRunCtxOutputTensor(pyg->g, i));
+        pyt->t = RedisAI_TensorGetShallowCopy(RedisAI_ModelRunCtxOutputTensor(mctx, i));
```
MeirShpilraien, line 3045: Don't you need to free mctx in the end?

src/redisgears_python.c
```diff
@@ -3025,15 +3040,83 @@ static PyObject* modelRunnerRun(PyObject *cls, PyObject *args){
...
+		const char *errStr = RedisAI_GetError(error);
+		PyObject* pyErr = PyUnicode_FromStringAndSize(errStr, strlen(errStr));
+		PyTuple_SetItem(pArgs, 1, pyErr);
+		PyObject_CallObject(setFutureExceptionFunction, pArgs);
+		goto finish;
+	}
+
+	for(size_t i = 0 ; i < RedisAI_ModelRunCtxNumOutputs(mctx) ; ++i){
+		PyTensor* pyt = PyObject_New(PyTensor, &PyTensorType);
+		pyt->t = RedisAI_TensorGetShallowCopy(RedisAI_ModelRunCtxOutputTensor(mctx, i));
```
MeirShpilraien, line 3074: Here also don't you need to free mctx?

src/redisgears_python.c
```diff
@@ -3013,10 +3022,16 @@ static PyObject* modelRunnerRun(PyObject *cls, PyObject *args){
...
     RAI_Error* err;
     RedisAI_InitError(&err);
+    RAI_ModelRunCtx *mctx = pyg->g;
+    pyg->g = NULL;
     PyThreadState* _save = PyEval_SaveThread();
-    RedisAI_ModelRun(&pyg->g, 1, err);
+    RedisAI_ModelRun(&mctx, 1, err);
     PyEval_RestoreThread(_save);
     if (RedisAI_GetErrorCode(err) != RedisAI_ErrorCode_OK) {
         PyErr_SetString(GearsError, RedisAI_GetError(err));
```
MeirShpilraien, line 3037: What if you return here? you still need to free mctx no?

MeirShpilraien (commented):
Last comment

MeirShpilraien (approved):

PR #476, opened 2021-01-12
Fixed RAMP file generation with Python plugin

rafie:


MeirShpilraien:
@rafie this fix changes the command line args to : `"command_line_args": "Plugin /home/meir/work/RedisGears/bin/linux-x64-debug/gears_python/gears_python.so"`
I guess we will not be able to fix it without adding support on ramp to setup different module args without modifying them in the final JSON artifact.

rafie:
@MeirShpilraien Good catch. I'll add the RAMP support.

MeirShpilraien:
Should only be merge after this : https://github.com/RedisLabs/RAMP/pull/60

MeirShpilraien (approved):

PR #514, opened 2021-02-17
Adding gmake4 to centos7

chayim:
The latest readies places a requirement on GNU Make >=4. This isn't yet wired up to CircleCI, but here so that I can find out:

1. Why the requirement on Make 4. Perhaps we don't need to enforce that and this can be erased.

2. If this is required, then stage the circleci change.

It should be noted that the changes in readies enforces Make 4, but RedisGraph builds with make 3.

### DIFF
--- a/system-setup.py
+++ b/system-setup.py
@@ -5,7 +5,8 @@ import os
 import argparse
 
 ROOT = HERE = os.path.abspath(os.path.dirname(__file__))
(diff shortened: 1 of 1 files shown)


chayim:
https://redislabs.slack.com/archives/G01Q8RN6J4R/p1614756869015400

PR #521, opened 2021-03-11
TensorGetValue APIs fix 

alonre24:
Following a change in RedisAI_TensorGetValueAsDouble/LongLong - fix the use in these APIs in PyTensor_ToFlatList: if it returns 0(=REDISMODULE_OK) it succeeded, and otherwise it failed (unlike before when it was the other way around).


MeirShpilraien:
@alon-brillix this will basically cause us not to be able to work with older versions of RedisAI, do we still have a release of RedisAI with this change? If we do we should check RedisAI version to decide how to handle this function return value. If there is not still a release with this change we might want to consider again breaking it...

alonre24:
Decided to remove the changes in RedisAI for not breaking Gears+RedisAI_1.0

PR #522, opened 2021-03-11
TensorGetValue APIs fix - for 1.0

alonre24:
Following a change in RedisAI_TensorGetValueAsDouble/LongLong in RedisAI - fix the use in these APIs in PyTensor_ToFlatList: if it returns 0(=REDISMODULE_OK) it succeeded, and otherwise it failed (unlike before when it was the other way around)


alonre24:
Decided to remove the changes in RedisAI for not breaking Gears+RedisAI_1.0

PR #529, opened 2021-03-23
[1.0] Build updates (incl. Redis 6.2)

rafie:


MeirShpilraien:
@rafie what will be the outcome of this PR? What cycles will run when?

rafie:
@MeirShpilraien This is basically corresponds to work that was done in master/1.2 (because it was impossible to cherry-pick), just to keep 1.0 close enough to master infrastructure-wise. I don't think we run any cycles on 1.0 now, apart from the usual builds when merging into it.

MeirShpilraien (approved):

PR #555, opened 2021-05-08
Avoid warnings from redisgears.h

rafie:


rafie:
See https://circleci.com/api/v1.1/project/github/RedisTimeSeries/RedisTimeSeries/8249/output/106/0?file=true&allocation-id=60950f5f8a4ccf017868faa4-0-build%2F4975710C

src/redisgears.h
```diff
@@ -9,8 +9,18 @@
 #define SRC_REDISGEARG_H_
 
 #include <stdbool.h>
+#include <limits.h>
 #include "redismodule.h"
-#include "utils/arr_rm_alloc.h"
+
```
MeirShpilraien, line 14: Maybe we should not do it in the code and put it on the Makefile?

src/common.h
```diff
@@ -13,6 +13,7 @@
 #include "cluster.h"
 #include <stdio.h>
 #include <stdbool.h>
+#include "utils/arr_rm_alloc.h"
```
MeirShpilraien, line 16: Why is it needed here?

MeirShpilraien (commented):
Some small questions

src/common.h
rafie: It's a quite common header, so instead of adding it to all source files that need it, I just inserted it into `common.h` (which is included into all .c files by a build flag).

src/redisgears.h
rafie: The point is that this file is to be included by a client code, so we don't really have a good control of how their Makefile looks like, and it will generate lots of warnings without the pragmas. Notice that I use push/pop, so the change is local to the part that defines the API functions.

MeirShpilraien: So why not go with the Redis approach and allow setting prefix and suffix to the function pointer def, this way you can add the 'extern' to each function and I believe it will solve it no?
https://github.com/redis/redis/blob/57b94eacaa64591295c9f9f9bb50eed9ed8b9a54/src/redismodule.h#L579

rafie: Well, I was trying to do something cosmetic that would not actually modify the API.
I basically agree, and also prefer our X-macro variant we failed to implement into OSS Redis.
And if we opened it up, why not take the approach we employed in Search/JSON, with a API of struct-of-pointers that will be made available to the client via GetSharedAPI?

MeirShpilraien: @rafie yes, I agree it should be a struct of pointers but this is a bigger change..

rafie: @MeirShpilraien I prefer to push it the way it is now (simply to be in sync with what RTS has now), and to arrange the API in the proper way, without going through an intermediate step. I can ask Omer to lend a hand.

MeirShpilraien (approved):

PR #565, opened 2021-05-31
changing dockers to apt update latest

chayim:


rafie:
@chayim LGTM. Is this for security patches?

chayim:
Yeah, by doing the default apt up, we get whatever is in the feeds.

rafie (approved):

Dockerfile
```diff
@@ -32,6 +32,10 @@ ADD . /build
 
 RUN ./deps/readies/bin/getpy2
 RUN ./system-setup.py
+RUN apt-get update -qq
```
rafie, line 35: @chayim This won't work on OSes that are not Debian-like. We better have something going for us within `system-sytem.py`.

chayim: Yeah, I just remembered that there are non-Ubuntu dockers... Though currently random. I'll re-inject support into Dockerfiles for updates via apt or yum for now. This isn't something I'd like system-setup to do, given that it also gets used on bare metal occasionally (heck, by myself).

rafie: I agree this is not `system-setup.py`'s role in life (though it's bound to do `apt-get update` to be able to install anything at all). We probably need to have this as a standard procedure in our Dockerfiles, just abstracted per package manager. So, how about something like `readies/bin/getget`, based on `is_command`, and simply performing the `apt-get upgrade` thing for each package manager. Maybe call it `getupdates`?

rafie:
Handled by #571, can close.

PR #569, opened 2021-06-13
Pytyhon compilation speedup for gears runs

### LINKED ISSUE #568: Test python compilation for speedups
Give this issue from [python.org](https://bugs.python.org/issue38980), it appears that we can get a speed boost in gears, by changing the compilation options.

chayim:
Changes to python compilation to (theoretically) speed up Gears runs, in the python interpreter. 

It's one thing to compile, but I'd like to qualify validating these. @filipecosta90 is there a way to do so?  The PR itself is here for approval, but I'd like to merge assuming we can validate this, rather than break things. WDYT? 
closes #568


MeirShpilraien:
@chayim is this still relevant?

chayim:
@filipecosta90 did you ever try this with perf?

chayim:

PR #578, opened 2021-07-08
Add null record type

danni-m:


src/record.h
```diff
@@ -76,14 +76,16 @@ extern RecordType* doubleRecordType;
...
 extern RecordType* asyncRecordType;
 extern Record StopRecord;
 extern Record WaitRecord;
 extern Record DummyRecord;
+extern Record NullRecord;
 
 Record* RG_GetDummyRecord();
 
-#define IS_SPECIAL_RECORD(r) (r == &StopRecord || r == &WaitRecord || r == &DummyRecord)
+#define IS_SPECIAL_RECORD(r) (r == &StopRecord || r == &WaitRecord || r == &DummyRecord || r == &NullRecord)
```
MeirShpilraien, line 88: I do not think NullRecord should be a special record...

src/record.c
```diff
@@ -12,6 +12,7 @@ RecordType StopRecordType;
 Record StopRecord;
 Record WaitRecord;
 Record DummyRecord;
+Record NullRecord;
```
MeirShpilraien, line 15: I do not see where you initialize it, what promise in the if I will do GetType on NullRecord I will get nullRecordType?

MeirShpilraien (commented):
:+1: Some comments

src/record.c
danni-m: @MeirShpilraien its a static var, do we need to initialise it? if so how?

MeirShpilraien (approved):

MeirShpilraien (approved):

PR #580, opened 2021-07-14
Removing pre-defined compiler settings (gcc)

chrisAtRedisLabs:
Using gcc (as in: the default version which is 10) on Hirsute breaks the build due to linker errors when compiling the final module. The RC for this a compiler standard for symbol visibility, using gcc-9 (make CC=gcc-9) is a work-around. Regression on hirsute with this work-around checks out.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -90,7 +90,6 @@ include build/hiredis/Makefile.defs
 
 #----------------------------------------------------------------------------------------------
 
-CC=gcc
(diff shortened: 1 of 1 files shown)


chayim:
It would be better to work with gcc 10. At the very least make CC=/path/to/your/gcc can be specified to override.

chayim:

PR #581, opened 2021-07-14
Update Redis Module API; support for GCC 10 and Ubuntu/Hirsute

rafie:


src/cluster.h
```diff
@@ -5,11 +5,9 @@
  *      Author: meir
  */
 
-#ifndef SRC_CLUSTER_H_
-#define SRC_CLUSTER_H_
+#pragma once
```
MeirShpilraien, line 8: Is it mandotory?

src/cluster.h
```diff
@@ -36,5 +34,7 @@ int Cluster_RefreshCluster(RedisModuleCtx *ctx, RedisModuleString **argv, int ar
 int Cluster_ClusterSet(RedisModuleCtx *ctx, RedisModuleString **argv, int argc);
 int Cluster_ClusterSetFromShard(RedisModuleCtx *ctx, RedisModuleString **argv, int argc);
 
-
-#endif /* SRC_CLUSTER_H_ */
+#define VERIFY_CLUSTER_INITIALIZE(c) \
```
MeirShpilraien, line 37: Was it mandatory to move it on this PR?

src/cluster.h
rafie: No, it simply looks nicer.

src/common.h
```diff
@@ -22,43 +19,9 @@
 #define MIN(a,b) (((a)<(b))?(a):(b))
 #define MAX(a,b) (((a)>(b))?(a):(b))
 
-extern Gears_dictType* dictTypeHeapIdsPtr;
```
MeirShpilraien, line 25: Was it mandatory to move those in this PR?

src/module.h
```diff
@@ -0,0 +1,15 @@
+
+#pragma once
```
MeirShpilraien, line 2: Why do we need this new file?

src/redisgears.h
```diff
@@ -5,8 +5,15 @@
...
-#define SRC_REDISGEARG_H_
+#pragma once
+
+#ifdef GEARS_API_MAIN
+#define GEARS_API
+#endif
+
+#ifndef GEARS_API
+#define GEARS_API extern
+#endif
```
MeirShpilraien, line 16: If we are already doing this change then why not also add a suffix like the redismodule.h?

MeirShpilraien (commented):
Some comments, basically I am not sure it's mandatory to move stuff between files, I believe that if it's not we should not do it in this PR.

PR #582, opened 2021-07-19
Update Redis Module API; support for GCC 10 and Ubuntu/Hirsute (2)

rafie:


src/version.h
```diff
@@ -35,4 +36,26 @@
 
 #define REDISGEARS_MODULE_NAME "rg"
 
-#endif /* SRC_VERSION_H_ */
+typedef struct RedisVersion{
```
MeirShpilraien, line 39: Why this was moved here?

src/cluster.h
```diff
@@ -36,5 +32,7 @@ int Cluster_RefreshCluster(RedisModuleCtx *ctx, RedisModuleString **argv, int ar
 int Cluster_ClusterSet(RedisModuleCtx *ctx, RedisModuleString **argv, int argc);
 int Cluster_ClusterSetFromShard(RedisModuleCtx *ctx, RedisModuleString **argv, int argc);
 
-
-#endif /* SRC_CLUSTER_H_ */
+#define VERIFY_CLUSTER_INITIALIZE(c) \
```
MeirShpilraien, line 35: Why this was moved here?

src/common.h
```diff
@@ -5,16 +5,20 @@
  *      Author: root
  */
 
-#ifndef SRC_COMMON_H_
-#define SRC_COMMON_H_
+#pragma once
```
MeirShpilraien, line 8: Do we must use this?

src/common.h
```diff
@@ -24,33 +28,10 @@
 
 extern Gears_dictType* dictTypeHeapIdsPtr;
 
-typedef struct RedisVersion{
```
MeirShpilraien, line 27: Why moving this?

src/global.c
```diff
@@ -0,0 +1,6 @@
+
+#define REDISMODULE_MAIN
```
MeirShpilraien, line 2: Why do we need this file?

src/version.h
rafie: It's version-related, seems logical. Do you prefer it like it was?

src/cluster.h
rafie: It's cluster-related, seems logical. Do you prefer it like it was?

src/common.h
rafie: It's more elegant (saves the bottom #endif). Any reason against using this?

src/global.c
rafie: `global.c` is excluded in Makefile from the `-include $(SRCDIR)/common.h`, therefore allowing pre-defining `REDISMODULE_MAIN`.

src/version.h
MeirShpilraien: I see this file more as holding constants of the RedisGears version. This is used as Redis version and potentially more commonly used, this is why I believe it should be on common.

src/cluster.h
MeirShpilraien: I agree, we can leave it here.

src/common.h
MeirShpilraien: Is it supported by all C compilers (or at least by the popular ones)?

src/redismodule.h
```diff
@@ -447,267 +580,306 @@ typedef struct RedisModuleTypeMethods {
...
-#define RedisModule_IsAOFClient(id) ((id) == CLIENT_ID_AOF)
+#define RedisModule_IsAOFClient(id) ((id) == UINT64_MAX)
 
 // enterprise api
-int REDISMODULE_API_FUNC(RedisModule_AvoidReplicaTraffic)();
-int REDISMODULE_API_FUNC(RedisModule_ShardingGetKeySlot)(RedisModuleString *keyname);
-void REDISMODULE_API_FUNC(RedisModule_ShardingGetSlotRange)(int *first_slot, int *last_slot);
+REDISMODULE_API int (*RedisModule_AvoidReplicaTraffic)() REDISMODULE_ATTR;
+REDISMODULE_API int (*RedisModule_ShardingGetKeySlot)(RedisModuleString *keyname) REDISMODULE_ATTR;
+REDISMODULE_API void (*RedisModule_ShardingGetSlotRange)(int *first_slot, int *last_slot) REDISMODULE_ATTR;
```
MeirShpilraien, line 879: You did not add those to REDISMODULE_GET_API

src/common.h
rafie: gcc, clang, msc and intel (same as gcc) support it.

src/version.h
rafie: no problem, will move it back.

MeirShpilraien (approved):

PR #587, opened 2021-08-21
added inorder option to command reader

MeirShpilraien:
- [x] Add test to check loading such registration from rdb


plugins/python/redisgears_python.c
```diff
@@ -1555,6 +1556,19 @@ static void* registerCreateCommandArgs(PyObject *kargs){
...
+    PyObject* pyInOrder = GearsPyDict_GetItemString(kargs, "inorder");
+    if(pyInOrder){
+        if(!PyBool_Check(pyInOrder)){
+            PyErr_SetString(GearsError, "inorder argument is be boolean");
+            return NULL;
+        }
+        if(Py_True == pyInOrder){
+            inOrder = 1;
+        }else{
+            inOrder = 0;
```
rafie, line 1568: `inOrder = Py_True == pyInOrder ? 1 : 0;` ?

docs/readers.md
```diff
@@ -342,12 +342,13 @@ The reader returns a list record with elements being the trigger's name followed
...
 
 ```python
-class GearsBuilder('CommandReader').register(trigger=None)
+class GearsBuilder('CommandReader').register(trigger=None, inorder=True/False)
 ```
 
 _Arguments_
 
 * _trigger_: the trigger's name
+* _inorder_: if `True`, the commands will run one after the other in the order they came (this option is only relevant for [`async_local`](functions.md#register) executions, on [`sync`](functions.md#register) executions you get it by default from the sync property, and its currently not possible to promise this property on [`async`](functions.md#register) executions ). Supported only on v1.0.7 and above.
```
rafie, line 351: The comment is not so clear - let's see how we can improve it.

rafie (approved):

rafie (approved):

PR #603, opened 2021-09-06
added python profiler when ProfileExecutions is enabled

MeirShpilraien:
- [x] tests
- [x] documentation

### DIFF
--- a/plugins/python/GearsBuilder.py
+++ b/plugins/python/GearsBuilder.py
@@ -1,7 +1,8 @@
 import redisgears
 import copy
 import redisgears as rg
-import cProfile, pstats
+import cProfile
+import pstats
 import io
 from redisgears import executeCommand as execute
 from redisgears import callNext as call_next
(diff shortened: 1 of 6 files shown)


rafie (approved):

plugins/python/GearsBuilder.py
```diff
@@ -1,6 +1,8 @@
 import redisgears
 import copy
 import redisgears as rg
+import cProfile, pstats
```
chayim, line 4:
```
import cProfile
import pstats
```

MeirShpilraien:
After @gkorland review, unified API to a single RG.PYPROFILE command with multiple subcommands.

plugins/python/GearsBuilder.py
MeirShpilraien: Fixed

MeirShpilraien:
API was changed according to @gkorland suggestion.

rafie (approved):

PR #608, opened 2021-09-22
Docker/debian: moved from Buster to Bullseye

rafie:


MeirShpilraien:
@rafie will this change affects any of the names of the artifacts?

rafie:
@MeirShpilraien We only have Ubuntu and CentOS artifacts on RedisGears, so no artifact names will be affected.

MeirShpilraien (approved):

PR #613, opened 2021-09-27
fix issue where stream reader might trigger executions on replica

MeirShpilraien:
- [ ] cherrypick to 1.0

### DIFF
--- a/pytest/test_register.py
+++ b/pytest/test_register.py
@@ -1427,3 +1427,36 @@ GB('CommandReader').foreach(SleepIfNeeded).map(lambda x: execute('lpush', x[1], 
         env.expect('lrange', 'l', '0', '-1').equal(['y', 'x'])
 
         env.cmd('flushall')
+
+def testStreamReaderNotTriggerEventsOnReplica(env):
+    env.skipOnCluster()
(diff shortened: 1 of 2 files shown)


pytest/test_register.py
```diff
@@ -1427,3 +1427,36 @@ def RunTestInOrderY():
         env.expect('lrange', 'l', '0', '-1').equal(['y', 'x'])
 
         env.cmd('flushall')
+
+def testStreamReaderNotTriggerEventsOnReplica(env):
+    env.skipOnCluster()
+    env.expect('RG.PYEXECUTE', "GB('StreamReader').map(lambda x: x['error']).register(onFailedPolicy='retry', onFailedRetryInterval=1)").equal('OK')
+    env.expect('xadd', 's', '*', 'foo', 'bar')
+    
+    res1 = env.cmd('RG.DUMPREGISTRATIONS')[0][7][3]
```
chayim, line 1436: Suggestion - move the command name into a variable and reuse. Otherwise fun.

MeirShpilraien: This is how we do it on all the tests, I would like to stay consistent.

chayim: Bummer. In redisplus we're consting all those strings. I like the pattern, it came in from django with their tuple-choices and now use it everywhere.

chayim (approved):

PR #631, opened 2021-11-11
[1.0] Build & CI updates

rafie:


chayim:
@rafie Can we reduce this to just what's needed for RHEL support? The docker changes for buster (etc) aren't related to RHEL and shouldn't be part of this commit.

MeirShpilraien:
Also, the changes to redisgears.h and redismodule.h, do we need it?

rafie:
@MeirShpilraien @chayim Everything in this PR are needed.

MeirShpilraien:
Is this still relevant?

PR #636, opened 2021-11-16
centos8

### LINKED ISSUE #427: Race condition on message passing can cause execution to reach timeout
When a node first connected to another node it sends internal RG.HELLO command, until the RG.HELLO reply reached the shard buffer all the messages that need to be sent to the other shard. When the RG.HELLO reply arrives the buffered messages are not sent. This causes the execution to wait until timeout. I believe that these occasional failures are happened because of this issue.
https://app.circleci.com/pipelines/github/RedisGears/RedisGears/2141/workflows/6ce4536d-6308-44ac-8039-3e64e199cd64/jobs/5114/steps

chayim:


Dockerfile
```diff
@@ -31,6 +31,7 @@ COPY --from=redis /usr/local/ /usr/local/
 ADD . /build
 
 RUN ./deps/readies/bin/getpy2
+RUN test -f /usr/bin/python || ln -s /usr/bin/python2 /usr/bin/python
```
rafie, line 34: Also possible: `FIX=1 ./deps/readies/bin/getpy2`

Makefile
```diff
@@ -117,7 +117,9 @@ OBJECTS=$(patsubst $(SRCDIR)/%.c,$(BINDIR)/%.o,$(SOURCES))
 CC_DEPS = $(patsubst $(SRCDIR)/%.c, $(BINDIR)/%.d, $(SOURCES))
 
 CC_FLAGS += \
-	-fPIC -std=gnu99 \
+	-fPIC \
+	-fcommon \
```
rafie, line 121: IMO this is more risky than my change, but I'll go with it.

MeirShpilraien: @rafie can you explain why? What risks do you see with it?

Dockerfile
chayim: We're not merging this back to master. It's a cut-throat patch. Didn't know, thanks - will use next time!

MeirShpilraien (approved):

PR #637, opened 2021-11-16
Used python decorator to perform cleanups after tests

MeirShpilraien:
Cleanup performed:
1. deleted all registration
2. delete all executions

This should cleanup occasional valgrind failure and in the future will allow us to run the tests with env-reuse flag and speed them up even more.

### DIFF
--- a/pytest/common.py
+++ b/pytest/common.py
@@ -1,6 +1,41 @@
 import signal
 from includes import *
 from threading import Thread
+from RLTest import Env
+import inspect
+
+class Colors(object):
+    @staticmethod
+    def Cyan(data):
+        return '\033[36m' + data + '\033[0m'
+
+    @staticmethod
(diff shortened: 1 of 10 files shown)


MeirShpilraien:
Require this to be merged: https://github.com/RedisLabsModules/RLTest/pull/150

pytest/common.py
```diff
@@ -1,6 +1,42 @@
 import signal
 from includes import *
 from threading import Thread
+from RLTest import Env, Defaults
+import inspect
+import unittest
+
+class Colors(object):
```
chayim, line 8: Maybe you want this library: https://pypi.org/project/colored/

MeirShpilraien: Lets do it on separate PR as RLTest also uses it.

chayim (approved):

PR #639, opened 2021-11-17
Upgrade pytest version to 6.2.5

alonre24:
After this github action error: https://github.com/RedisGears/RedisGears/runs/4234637668?check_suite_focus=true
Try to solve this based on: https://stackoverflow.com/questions/69528842/an-error-while-trying-to-execute-tests-on-python-3-10-with-pytest


chayim (approved):
YAY!

PR #641, opened 2021-11-18
Add redisai docs to nav bar and home page

alonre24:
After uploading RedisAI docs, I noticed that it is not accessible from the website. This PR should address this.


itamarhaber (approved):
LGTM

PR #647, opened 2021-12-15
Docker size change to test valgrind times

chayim:


MeirShpilraien:
@chayim this did not reduce the time, its just failed on the 2 shards tests so it did not run the 3 shards. This should reduce the time once I finalise it: https://github.com/RedisGears/RedisGears/pull/637

chayim:
Right - this was purely a quick test for the win. When ready from your end, let me know and we'll see what actually happens.

MeirShpilraien:
@chayim I guess we can close right?

PR #661, opened 2022-01-11
Fix #536, added RG.TRIGGERONKEY

### LINKED ISSUE #536: Fix RG.TRIGGER third argument to be a key on oss cluster
RG.TRIGGER command is not declared correctly, the third argument should be considered a key:
https://github.com/RedisGears/RedisGears/blob/ed10294edb4ff8e06f9b53c03b6d730b0377e82d/src/readers/command_reader.c#L1030

MeirShpilraien:
The **RG.TRIGGERONKEY** command is the same as **RG.TRIGGER**, the difference is that the third argument is considered a key so client will know how to direct the command to the correct shard.

**Redis API**

```
RG.TRIGGERONKEY <trigger> key [arg ...]
```

_Arguments_

* **trigger**: the trigger's name
* **key**: a key on which the trigger is executed on
* **arg**: any additional arguments

_Return_

An array containing the function's output records.

**Examples**

```
redis> RG.PYEXECUTE "GB('CommandReader').map(lambda x: execute('set', x[1], x[2])).register(trigger='my_set')"
OK
redis> RG.TRIGGERONKEY my_set foo bar
1) "OK"
redis> get foo
bar
```


src/readers/command_reader.c
```diff
@@ -1086,7 +1086,12 @@ int CommandReader_Initialize(RedisModuleCtx* ctx){
     // using rm_call. In this case the effect of the execution is replicated
     // and not the execution itself.
     if (RedisModule_CreateCommand(ctx, GEARS_OVERRIDE_COMMAND, CommandReader_Trigger, "readonly", 0, 0, 0) != REDISMODULE_OK) {
-        RedisModule_Log(staticCtx, "warning", "could not register command rg.command");
+        RedisModule_Log(staticCtx, "warning", "could not register command "GEARS_OVERRIDE_COMMAND);
```
chayim, line 1089: "could not register command "GEARS_OVERRIDE_COMMAND 
^^ seems off

MeirShpilraien: Why? we try to register it and failed.

pytest/test_async.py
```diff
@@ -1038,10 +1038,22 @@ async def doTest(x):
...
+    wait_for_execution = True
+    with TimeLimit(5, env, 'Failed waiting for execution to start running'):
+        while wait_for_execution:
+            executions = env.cmd('RG.DUMPEXECUTIONS')
+            for e in executions:
+                if e[3] == 'running':
+                    wait_for_execution = False
+                    env.expect('RG.ABORTEXECUTION', e[1]).equal('OK')
+                    break
+                if e[3] == 'done':
```
chayim, line 1051: elif

MeirShpilraien: Thanks, fixed.

chayim:
@MeirShpilraien last thing - shouldn't we add to commands.json. Approved otherwise.

chayim (approved):

PR #667, opened 2022-01-13
Use safe lock in RedisAI initialization

alonre24:


MeirShpilraien (approved):
@alonre24 do we need to backport it to 1.0? If so please open a PR.

alonre24:
> @alonre24 do we need to backport it to 1.0? If so please open a PR.
As far as I understand we don't need, since in 1.0 RedisAI plugin was initialized with MODULE_LOADED event (without locking the GIL explicitly).

MeirShpilraien:
So why it was changed?

alonre24:
good question. @DvirDukhan Do you remember?

DvirDukhan:
I believe it was part of #461 
moved from `RedisGears_OnModuleLoad` to redisgears_python
@MeirShpilraien thought?

MeirShpilraien:
Ohh right I remembered now, we need to expose this API to plugin.

PR #676, opened 2022-01-19
Fix globals dictionary not set correctly after deserialization.

MeirShpilraien:
The problem
---

After deserialization, the deserialized function are created with its own globals dictionary. We want all the functions on the same execution (and even different execution that was generated by the same `RG.PYEXECUTE` command) to share the same globals dictionary. The current solution is to change the function global dictionary to be the global dictionary of the `PY.EXECUTE` session that created the execution.

This solution only changes the top most function, but it does not change function that is used by it, so the following example will break:

```
g = 1
def f1():
    global g
    g = g + 1
    print(id(globals()))
    return g

def f(x):
    global g
    g = g + 1
    print(id(globals()))
    return f1()

GB('CommandReader').map(f).register(trigger='test', convertToStr=False)
GB('CommandReader').map(f).register(trigger='test1', convertToStr=False)
```

Running `RG.TRIGGER test` and `RG.TRIGGER test1` will give the following:

```
> RG.TRIGGER test
3
> RG.TRIGGER test1
5
```

But after RDB load we will get the following result (that proves that `f1` was created with a different global dictionary):

```
> RG.TRIGGER test
3
> RG.TRIGGER test1
4
```

The solution
---

Modify `cloudpickle` to look for a global variable, `requested_base_globals`. If set, use this dictionary as the global dictionary of all the created functions. RedisGears will make sure to set `requested_base_globals` to the correct value before deserializing a function. In addition, changed `cloudpickle` to update the global dictionary only with the missing fields and not override existing fields. After this change, the above example works as expected.

Test was added to verify the fix.

### DIFF
--- a/plugins/python/cloudpickle.py
+++ b/plugins/python/cloudpickle.py
@@ -1054,7 +1054,10 @@ def _fill_function(*args):
     #   content of state['globals'] to populate the shared isolated
     #   namespace with all the global variables that are specifically
     #   referenced for this function.
-    func.__globals__.update(state['globals'])
+    #   update only keys that does not exists in the current global dictionary
+    for k,v in state['globals'].items():
+        if k not in func.__globals__.keys():
+            func.__globals__[k] = v
 
     func.__defaults__ = state['defaults']
     func.__dict__ = state['dict']
@@ -1086,7 +1089,7 @@ def _make_empty_cell():
 
     return (lambda: cell).__closure__[0]
 
-
+requested_base_globals = None
 def _make_skel_func(code, cell_count, base_globals=None):
     """ Creates a skeleton function object that contains just the provided
         code and the correct number of cells in func_closure.  All other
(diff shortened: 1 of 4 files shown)


plugins/python/cloudpickle.py
```diff
@@ -1054,7 +1054,10 @@ def _fill_function(*args):
     #   content of state['globals'] to populate the shared isolated
     #   namespace with all the global variables that are specifically
     #   referenced for this function.
-    func.__globals__.update(state['globals'])
+    #   update only keys that does not exists in the current global dictionary
+    for k,v in state['globals'].items():
+        if k not in func.__globals__.keys():
```
chayim, line 1059: Can you use func.__globals__,update(state['globals']) here?

MeirShpilraien: This is how it was before, but updates override collides keys and this is not what I want.

chayim: Correct - it would override all - that's why I was askling if it's different. Cool. approved.

chayim (approved):

chayim (approved):

PR #686, opened 2022-01-24
Adding JVM OS as a variable to pull down openjdk

chayim:
- snapshots do not depend on osx now
- jvmos for osx and linux, validating


MeirShpilraien:
@chayim is this needed?

chayim:
No longer necessary.

PR #707, opened 2022-02-01
fix missing JVM_SessionDup

MeirShpilraien:


chayim (approved):

chayim (approved):
Fixed the error string too I see

chayim (approved):

PR #718, opened 2022-02-07
fix mac build

MeirShpilraien:


Makefile
```diff
@@ -418,6 +418,12 @@ ifeq ($(VALGRIND),1)
 TEST_FLAGS += VALGRIND=1
 endif
 
+ifeq ($(OS),macos)
+PARALLELISM=1
```
chayim, line 422: That's sad if you think about it.

plugins/jvmplugin/Makefile
```diff
@@ -14,9 +14,19 @@ $(error Build redisgears first)
 endif
 endif
 
+ifeq ($(OSNICK),catalina)
+JVM_DOWNLOAD_LING=https://github.com/adoptium/temurin11-binaries/releases/download/jdk-11.0.14%2B9/OpenJDK11U-jdk_x64_mac_hotspot_11.0.14_9.tar.gz
```
chayim, line 18: Nit: JVM_DOWNLOAD_LING should be JVM_DOWNLOAD_LINK

chayim (approved):

PR #721, opened 2022-02-09
Test with Redis 7.0-rc1

gkorland:


chayim:
No longer necessary in the 7.2x world.

PR #723, opened 2022-02-13
Locale change, as a result of the build error

chayim:
The build used to error as below. This locale change should solve that.

```
This system supports the C.UTF-8 locale which is recommended. You might be able to resolve your issue by exporting the following environment variables:

    export LC_ALL=C.UTF-8
    export LANG=C.UTF-8
```

### DIFF
--- a/pack.sh
+++ b/pack.sh
@@ -11,6 +11,7 @@ export READIES=$ROOT/deps/readies
 cd $ROOT
 
 export PYTHONWARNINGS=ignore
+export LC_CTYPE=en_US.UTF-8
 
 #----------------------------------------------------------------------------------------------
 


MeirShpilraien (approved):

rafie:
@chayim This should be handled by `readies/bin/enable-utf8` and not from within the script (as this repeat itself ad infinitum and we better handle it globally).

chayim:
> @chayim This should be handled by `readies/bin/enable-utf8` and not from within the script (as this repeat itself ad infinitum and we better handle it globally).

Doing so is more destructive. It modifies the environment permanently by writing a script to disk - meaning many users (like myself) can't run this script given our environments.

PR #734, opened 2022-02-27
Pause/Unpause registrations

### LINKED ISSUE #722: RG.DUMPREGISTRATIONS - onFailedRetryInterval argument is missing in args section
In dumpregistrations, I'd expect to get the onFailedRetryInterval too, got this:
  ```
23) "args"
       24) 1) "batchSize"
           2) (integer) 2000
           3) "durationMS"
           4) (integer) 10000
           5) "stream"
           6) "*"
```

### LINKED ISSUE #724: Pause/unpause gear function
Would be nice to support this instead of unregistering (to simulate stop operation) the gear function (with abort pending or without it) and reregistering it (to simulate start operation)

### PR DESCRIPTION
**Pull request by MeirShpilraien**
Fix #724, add support for registrations pause and unpause.
The PR introduce 2 new commands:

* `RG.PAUSEREGISTRATIONS [ABORTPENDINGS] <ID1> [<ID2> ...]`
* `RG.UNPAUSEREGISTRATIONS <ID1> [<ID2> ...]`

The new commands allows to pause/unpause registration from triggering new executions. Currently, only stream registrations support pausing/unpausing. An attempt to pause/unpause a registration that do not use the stream reader
will raise an error.

It is also possible to unpause a registration from within the registration code by raising a special error containing the `PAUSE` string, example:

```python
GB('StreamReader').foreach(lambda x: flatError('PAUSE registration is paused')).register()
```

pause/unpause a registration that already pause/unpause is consider as no op and will keep the state exactly as it is (without any errors).

pause/unpause is considered an atomic operation, all the registrations that was given will be pause/unpause together and if one failed, the entire operation is aborted (atomicity is promised on the shard level and not on the cluster level).

Unpausing a registration will restart processing data from the stream, **if the stream is not set to trim messages (using `trimStream` option) then all the element will potentially be processed again**.

This PR also fix #722, added `trimStream`, `onFailedPolicy` and `retryInterval` to `RG.DUMPREGISTRATIONS`.

todo:
- [x] Update docs
- [x] Allow pause execution by returning a special error type
- [x] Testing

### DIFF
--- a/plugins/python/GearsBuilder.py
+++ b/plugins/python/GearsBuilder.py
@@ -171,16 +171,16 @@ class GearsBuilder():
             arg = createKeysOnlyReader(arg, **kargs)
         return self.gearsCtx.run(arg, **kargs)
 
-    def register(self, prefix='*', convertToStr=True, collect=True, **kargs):
+    def register(self, prefix='*', convertToStr=True, collect=True, **kwargs):
         if(convertToStr):
             self.gearsCtx.map(lambda x: str(x))
         if(collect):
             self.gearsCtx.collect()
-        kargs['prefix'] = prefix # this is for backword comptability
-        if 'regex' in kargs:
+        kwargs['prefix'] = prefix # this is for backword comptability
+        if 'regex' in kwargs:
             log('Using regex argument with register is deprecated and missleading, use prefix instead.', level='warning')
-            kargs['prefix'] = kargs['regex']
-        return self.gearsCtx.register(**kargs)
+            kwargs['prefix'] = kwargs['regex']
+        return self.gearsCtx.register(**kwargs)
 
 def createDecorator(f):
     def deco(self, *args):
--- a/plugins/python/GearsBuilder.py
+++ b/plugins/python/GearsBuilder.py
@@ -21,7 +21,7 @@ from redisgears import isInAtomicBlock
 from redisgears import config_get as configGet
 from redisgears import PyFlatExecution
 from redisgears import isAsyncAllow as isAsyncAllow
-from redisgears import flatError as flatError
+from redisgears import flatError as flat_error
 import asyncio
 from asyncio.futures import Future
 from threading import Thread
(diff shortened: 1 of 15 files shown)


plugins/python/GearsBuilder.py
```diff
@@ -179,7 +180,7 @@ def register(self, prefix='*', convertToStr=True, collect=True, **kargs):
         if 'regex' in kargs:
             log('Using regex argument with register is deprecated and missleading, use prefix instead.', level='warning')
             kargs['prefix'] = kargs['regex']
-        self.gearsCtx.register(**kargs)
+        return self.gearsCtx.register(**kargs)
```
chayim, line 183: Nit: kargs should be kwargs

pytest/test_register.py
```diff
@@ -1751,3 +1751,168 @@ def test(x):
...
+    env.assertEqual(counterY, '1')
+
+    env.execute_command('RG.PAUSEREGISTRATIONS', regId)
+
+    conn.execute_command('xadd', 'x', '*', 'foo', 'bar')
+    conn.execute_command('xadd', 'y', '*', 'foo', 'bar')
+
+    counterX = conn.execute_command('get', 'counter{x}')
+    counterY = conn.execute_command('get', 'counter{y}')
+    env.assertEqual(counterX, '1')
```
chayim, line 1785: Can't all of these env.assertEqual be wrtten as (example):

assert counterX == '1'

MeirShpilraien: I prefer to keep the convention we already have

plugins/python/GearsBuilder.py
MeirShpilraien: Fixed

pytest/test_register.py
chayim: We should open another item to clean these and adopt modern asserts. Fair - out of scope.

chayim (approved):

PR #739, opened 2022-03-14
Added tls support

MeirShpilraien:
Require this fix: https://github.com/RedisLabsModules/RLTest/pull/160

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -195,6 +195,15 @@ jobs:
           test_params: VALGRIND=1
           getredis_params: --valgrind
 
+  tls:
+    docker:
+      - image: redisfab/rmbuilder:6.2.5-x64-bullseye
+    steps:
+      - build-steps:
+          platform: debian
+          build_params: TLS=1
+          test_params: TLS=1
+
   build-platforms:
     parameters:
       platform:
@@ -397,6 +406,8 @@ workflows:
               platform: [hirsute, focal, bionic, xenial, centos7, rocky8, bullseye]
       - valgrind:
           <<: *on-integ-and-version-tags
+      - tls:
+          <<: *on-any-branch
       - build-macos:
           <<: *on-integ-and-version-tags
       - jvmpublish:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -407,7 +407,7 @@ workflows:
       - valgrind:
           <<: *on-integ-and-version-tags
       - tls:
-          <<: *on-any-branch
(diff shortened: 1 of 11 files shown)


.circleci/config.yml
```diff
@@ -195,6 +195,15 @@ jobs:
           test_params: VALGRIND=1
           getredis_params: --valgrind
 
+  tls:
+    docker:
+      - image: redisfab/rmbuilder:6.2.5-x64-bullseye
```
chayim, line 200: Might as well use 6.2.6 if we're already here.

.circleci/config.yml
```diff
@@ -397,6 +406,8 @@ workflows:
               platform: [hirsute, focal, bionic, xenial, centos7, rocky8, bullseye]
       - valgrind:
           <<: *on-integ-and-version-tags
+      - tls:
+          <<: *on-integ-and-version-tags
```
chayim, line 410: Any reason we don't want this on all branches?

pytest/run_tests_tls.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+HERE="$(cd "$(dirname "${BASH_SOURCE[0]}")" >/dev/null 2>&1 && pwd)"
+
+$HERE/generate_tests_cert.sh
```
chayim, line 5:
```
${HERE}
```

pytest/run_tests_tls.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+HERE="$(cd "$(dirname "${BASH_SOURCE[0]}")" >/dev/null 2>&1 && pwd)"
```
chayim, line 3: Can't your just use HERE=`${BASH_SOURCE[0]}` or do you need the full working directory?

chayim (approved):
Small changes. I'm approving either way as they're not critical.

.circleci/config.yml
MeirShpilraien: Juts taking time and I do not think it is needed...

pytest/run_tests_tls.sh
MeirShpilraien: I prefer a full working folder.

.circleci/config.yml
MeirShpilraien: Fixed

pytest/run_tests_tls.sh
MeirShpilraien: Fixed

PR #741, opened 2022-03-16
Added build for python only artifacts

MeirShpilraien:


pack.sh
```diff
@@ -80,17 +80,27 @@ pack() {
...
 	local packfile=artifacts/$artifact/$pack_fname
-	${PYTHON_BIN} $READIES/bin/xtx \
-		-d GEARS_PYTHON_NAME=python3_$SEMVER \
-		-d GEARS_PYTHON_FNAME=$URL_FNAME \
-		-d GEARS_PYTHON_SHA256=$(cat $GEARSPY_PKG.sha256) \
-		-d GEARS_JAVA_FNAME=$JAVA_URL_FNAME \
-		-d GEARS_JAVA_SHA256=$(cat $GEARSJVM_PKG.sha256) \
-		-d OS_DESC=$OS_DESC \
-		ramp.yml > /tmp/ramp.yml
+	if [[ $python_only == "true" ]]; then
```
chayim, line 87: instead of true/false, maybe just if [[ -z $ 3 ]]

MeirShpilraien: Fixed

PR #756, opened 2022-04-18
Version jvmplugin from the version passed in

chayim:


ramp.yml
```diff
@@ -10,15 +10,14 @@ min_redis_pack_version: '6.0.12'
...
-    - crdb 
+    - crdb
     - failover_migrate
     - persistence_aof
     - persistence_rdb
     - clustering
     - backup_restore
     - reshard_rebalance
     - eviction_expiry
-    - flash
```
MeirShpilraien, line 21: If we are merging it to master, this should not be removed.

ramp_python.yml
```diff
@@ -18,7 +18,6 @@ capabilities:
     - backup_restore
     - reshard_rebalance
     - eviction_expiry
-    - flash
```
MeirShpilraien, line 21: Same

src/version.h
```diff
@@ -9,9 +9,9 @@
 
 #include <stdbool.h>
 
-#define REDISGEARS_VERSION_MAJOR 99
-#define REDISGEARS_VERSION_MINOR 99
-#define REDISGEARS_VERSION_PATCH 99
+#define REDISGEARS_VERSION_MAJOR 1
```
MeirShpilraien, line 12: Why changing the version on master?

MeirShpilraien (commented):
Some comments

MeirShpilraien (approved):

PR #762, opened 2022-04-28
Fix MOD-3054. On enterprise, if not plugin was specified, automatically load the python plugin.

MeirShpilraien:
Fix MOD-3054. On enterprise, if not plugin was specified, automatically load the python plugin.

### DIFF
--- a/src/module.c
+++ b/src/module.c
@@ -1289,41 +1289,54 @@ static int Command_NetworkTest(RedisModuleCtx *ctx, RedisModuleString **argv, in
     return REDISMODULE_OK;
 }
 
-static int RedisGears_InitializePlugins(RedisModuleCtx *ctx) {
-    char** plugins = GearsConfig_GetPlugins();
-    for(size_t i = 0 ; i < array_len(plugins) ; ++i){
-        char* plugin = plugins[i];
-        char* pluginFile = NULL;
-        const char* moduleDataDir = getenv("modulesdatadir");
-        if(moduleDataDir){
-            // modulesdatadir env var exists, we are running on redis enterprise and we need to run on modules directory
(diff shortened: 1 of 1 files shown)


src/module.c
```diff
@@ -1289,41 +1289,54 @@ static int Command_NetworkTest(RedisModuleCtx *ctx, RedisModuleString **argv, in
...
-            rg_asprintf(&pluginFile, "%s/%s/%d/deps/%s/plugin/%s.so", moduleDataDir, REDISGEARS_MODULE_NAME, REDISGEARS_MODULE_VERSION, plugin, plugin);
-        }else{
-            pluginFile = RG_STRDUP(plugin);
-        }
+static int RedisGears_LoadSinglePlugin(RedisModuleCtx *ctx, char* plugin) {
+    char* pluginFile = NULL;
+    const char* moduleDataDir = getenv("modulesdatadir");
+    if(moduleDataDir){
+        // modulesdatadir env var exists, we are running on redis enterprise and we need to run on modules directory
+        rg_asprintf(&pluginFile, "%s/%s/%d/deps/%s/plugin/%s.so", moduleDataDir, REDISGEARS_MODULE_NAME, REDISGEARS_MODULE_VERSION, plugin, plugin);
```
chayim, line 1297: Question - I know we're (hard) tied to unix paths already - but I make it a point in python to use os.path.join so that at leas tpaths are platform independent. Is there something similar we can do here?

MeirShpilraien: Don't think we support any OK which are not linux base, but I see you point. I am not sure there is something similar to os.path.join in C, maybe there is an extrnal library for that but not sure we want to use another library just for this. Will check it out...

MeirShpilraien: @chayim I think its better to leave it to another PR, its already all over the code, WDYT?

chayim: Yeah - sounds like we just generally should be OS agnostic rather than here. It's philosophy - I chip away, but we can do another. Stamped

chayim (approved):

PR #763, opened 2022-04-28
Fix #761, Fix #760, stream reader fixes to Pause and Unregister

### LINKED ISSUE #760: Unregistering  a StreamReader's function keeps processing the stream elements until it's empty
I'd expect it to just finish processing the current batch.

### LINKED ISSUE #761: When pausing a stream based registration, the current batch is processed but its records remain in the stream

MeirShpilraien:
Fix #761, Fix #760, stream reader fixes to Pause and Unregister
1. Fix stream reader pause to ack after finish processing the current batch.
2. Fix stream reader unregister to stop processing more batches when finished the current batch.

### DIFF
--- a/src/readers/streams_reader.c
+++ b/src/readers/streams_reader.c
@@ -84,7 +84,7 @@ typedef struct SingleStreamReaderCtx{
     long long pendingMessages;
     long long nextBatch;
     bool isRunning;
-    bool freeWhenDone;
+    bool isFreeWhenDone;
 }SingleStreamReaderCtx;
 
 static void* StreamReader_ScanForStreams(void* pd);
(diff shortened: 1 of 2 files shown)


src/readers/streams_reader.c
```diff
@@ -84,7 +84,7 @@ typedef struct SingleStreamReaderCtx{
     long long pendingMessages;
     long long nextBatch;
     bool isRunning;
-
+    bool freeWhenDone;
```
chayim, line 87: Nit: Maybe isFreeWhenDone to match isRunning?

MeirShpilraien: Fixed

chayim (approved):

PR #765, opened 2022-05-01
Fix #764, allow disabling costume python allocator using module argument.

### LINKED ISSUE #764: Worse performance when invoking the same code in a gear function comparing to a standalone python
Performed the same test on the same machine under the same venv that RedisGears uses and got lower performance.
Code snippet:
```
def bench(x):
    iterations = 200000

    pipeline_size = 1000

    r = redis.Redis(
        host="localhost",
        port=12000)
    p = r.pipeline()
    cmd = {}
    payload = "1234567890".encode()
    for i in range(50):
        cmd[f"keyname{i}".encode()] = payload

    start = time.time()
    for i in range(iterations//processes):
        if (i % pipeline_size == 0):
            p.execute()
            p = r.pipeline(transaction=False)
        if (i % 1000 == 0):
            print(".", end='', flush=True)
        p.hset(f"mykey{i%4}".encode(), mapping=cmd)
    p.execute()
    end = time.time()
    # print(f"total time {(end-start)*1000:.0f}, {iterations//processes/(end-start)}/sec")
    return orjson.dumps({"total": f"{(end-start)*1000:.0f}, {iterations//processes/(end-start)}/sec"}).decode("utf-8")

```
Results:
Via gear's ComamndReader: 
1) "{\"total\":\"81202, 2462.986792641561/sec\"}"

Standalon python:
total time 49893, 4008.5886531942483/sec

MeirShpilraien:
Fix #764, allow disabling custom python allocator using module argument.

Introduce a new module argument to disable Redis custom python allocator. The new module argument called `OverridePythonAllocators` and can be set to either `1` or `0`. Disabling custom python allocator allows to achieve better performance (in some cases we saw improvement of up to 50%), the disadvantage is that the report of `RG.PYSTATS` command becomes invalid and Redis will not be able to track and report the memory usage used by the python interpreter.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -120,7 +120,7 @@ _Runtime Configurability_
 Supported
 
 ## OverridePythonAllocators
-The **OverridePythonAllocators** configuration option controls whether RedisGears will override the default python memory allocators. By disabling this option the python interpreter can achieve better performance (in some cases we saw improvement of up to 50%), the disadvantage is that the report of [RG.PYSTATS](commands.md#rgpystats) command is no longer valid and Redis will not be able to track and report the memory usage used by the python interpreter.
+The **OverridePythonAllocators** configuration option controls whether RedisGears will override the default python memory allocators. Disabling this option, causes the python interpreter to increase performance (in some cases we saw improvement of up to 50%), the disadvantage is that the output of [RG.PYSTATS](commands.md#rgpystats) becomes invalid, and Redis will be unable to track and report memory usage by the python interpreter.
 
 !!! important "Notice"
     Available sense v1.2.4
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -119,6 +119,24 @@ _Runtime Configurability_
 
 Supported
 
+## OverridePythonAllocators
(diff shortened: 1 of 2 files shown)


docs/configuration.md
```diff
@@ -119,6 +119,24 @@ _Runtime Configurability_
 
 Supported
 
+## OverridePythonAllocators
+The **OverridePythonAllocators** configuration option controls whether RedisGears will override the default python memory allocators. By disabling this option the python interpreter can achieve better performance (in some cases we saw improvement of up to 50%), the disadvantage is that the report of [RG.PYSTATS](commands.md#rgpystats) command is no longer valid and Redis will not be able to track and report the memory usage used by the python interpreter.
```
chayim, line 123:
```
The **OverridePythonAllocators** configuration option controls whether RedisGears will override the default python memory allocators. Disabling this option, causes the python interpreter to increase performance (in some cases we saw improvement of up to 50%), the disadvantage is that the output of [RG.PYSTATS](commands.md#rgpystats) becomes invalid, and Redis will be unable to track and report memory usage by the python interpreter.
```

chayim (changes requested):
Just a docs thing.

docs/configuration.md
MeirShpilraien: Fixed

chayim (approved):

PR #776, opened 2022-05-15
Fix flaky tests by adding MaxIdle time to register session.

MeirShpilraien:
Requirement installation was moved to be part of the session update processes, we need to allow the upgrade processes to continue long enough without a timeout. This PR changes the code to use `PythonInstallReqMaxIdleTime` config value to set the session update MaxIdle time.

### DIFF
--- a/src/execution_plan.h
+++ b/src/execution_plan.h
@@ -295,6 +295,7 @@ typedef struct SessionRegistrationCtx {
     RegistrationData *registrationsData;
     Gears_Buffer* buff;
     Gears_BufferWriter bw;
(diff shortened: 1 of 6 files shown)


src/execution_plan.h
```diff
@@ -295,6 +295,7 @@ typedef struct SessionRegistrationCtx {
     RegistrationData *registrationsData;
     Gears_Buffer* buff;
     Gears_BufferWriter bw;
+    long long maxIdle;
```
chayim, line 298: long long?

chayim (approved):

PR #772, opened 2022-05-09
Added the ability to upgrade dependencies

MeirShpilraien:
Fix #443, Added the ability to force upgrade dependencies without a restart.

A new flag was added to `RG.PYEXECUTE` called `FORCE_REQUIREMENTS_REINSTALLATION`. The new flag indicate that a re-installation of the requirement is needed. In order to support requirement update without a restart, RedisGears will reset the python interpreter modules cache after the re-installation is finished. This way the modules will be reloaded by the python interpreter.

Limitations:
* It is not possible to have multiple version of the same requirement at the same time, if one registration used an old version of a requirement and another registration upgrade it, then after restart they will both have the new version and if the new version is not backward compatible then the code might break with errors.

todo:
- [x] Update docs
- [x] Improve tests

PR #779, opened 2022-05-15
docs fixes

MeirShpilraien:
According to the comments on: https://github.com/RedisGears/RedisGears/pull/772#discussion_r871384901


docs/commands.md
```diff
@@ -472,7 +472,7 @@ _Arguments_
...
+* _FORCE_REINSTALL_REQUIREMENTS_: Force re-installation of all the requirements. When this option is used RedisGears will reset the python interpreter modules cache after the re-installation is finished. This way the modules will be reloaded by the python interpreter. One important limitation of using this option is that **It is not possible to have multiple versions of the same requirement at the same time**, if one registration used an old version of a requirement and another registration upgrade it, then after restart (or even before restart on some rare cases) they will both have the new version and if the new version is **not** backward compatible **the code might break with errors**. This option is available since v1.2.4.
```
chayim, line 475: ```This option is available as of v1.2.4.```

chayim, line 475:
```
* _FORCE_REINSTALL_REQUIREMENTS_: Force re-installation of all the requirements. When this option is used, RedisGears will reset the python interpreter module cache after re-installation is finished. The python interpreter then reloads the modules. An important limitation of this option is that **It is not possible to have multiple versions of the same requirement at the same time**. If one registration uses an old version of a requirement and another registration upgrades it, then after a restart (or before, on some rare cases) both will have the new version. If the new version is **not** backwards compatible **the code may fail due to errors**. This option is available as of v1.2.4.
```

docs/configuration.md
```diff
@@ -123,7 +123,7 @@ Supported
 The **OverridePythonAllocators** configuration option controls whether RedisGears will override the default python memory allocators. Disabling this option, causes the python interpreter to increase performance (in some cases we saw improvement of up to 50%), the disadvantage is that the output of [RG.PYSTATS](commands.md#rgpystats) becomes invalid, and Redis will be unable to track and report memory usage by the python interpreter.
 
 !!! important "Notice"
-    Available sense v1.2.4
+    Available since v1.2.4
```
chayim, line 126:
```
    Available as of v1.2.4
```

chayim (approved):

PR #798, opened 2022-07-04
Fix #792, trigger execution infinitely when trimming is disabled.

### LINKED ISSUE #792: Registration keeps fire executions when trimming is disabled.
When trimming is disable, RedisGears consumer group might not be at the beginning of the stream, this will cause RedisGears to believe there are always more elements to processes and cause intinite loop of triggered executions. Possible solution is to randomly chose the consumer group name so it will always start from the beginning.

MeirShpilraien:
Fix #792, trigger execution infinitely when trimming is disabled.

When trimming is disabled, if the registration was registered on a stream such that gears consumer group are not point to the beginning of the stream, RedisGears would enter a state where it triggers execution infinitely. Such situation can happened when upgrade a registrations (that has trimming disable) or when loading from RDB (again when trimming disabled).

This happened because RedisGears decides to trigger executions base on the amount of pending data which is calculated by checking the stream length. This check is wrong when trimming is disabled because the stream length is not a good indication to the amount of pending data.

To solve this issue, RedisGears will use a different way to decide if there is more data to processes. If, in the current batch, there was no data to processes and no new data was added to the stream during the batch execution, we know there is
no more data to processes and we avoid trigger another execution.

### DIFF
--- a/pytest/test_register.py
+++ b/pytest/test_register.py
@@ -2082,12 +2082,12 @@ GB('StreamReader').foreach(lambda x: execute('incr', 'x')).register(batch=2, onF
             while True:
                 registrations = env.cmd('RG.DUMPREGISTRATIONS')
                 if registrations[0][7][3] > 2:
-                    self.env.assertTrue(False, message='More than 2 executions was triggered')
+                    self.env.assertTrue(False, message='More than 2 executions were triggered')
                     break
                 time.sleep(0.1)
     except Exception as e:
         if 'timeout' not in str(e):
-            print(str(e))
+            sys.stderr.write('%s\n' % str(e))
         env.assertContains('timeout', str(e))
 
     env.cmd('xadd', 's', '*', 'foo', 'bar')
@@ -2098,10 +2098,10 @@ GB('StreamReader').foreach(lambda x: execute('incr', 'x')).register(batch=2, onF
             while True:
                 registrations = env.cmd('RG.DUMPREGISTRATIONS')
                 if registrations[0][7][3] != 3:
-                    self.env.assertTrue(False, message='More than 2 executions was triggered')
+                    self.env.assertTrue(False, message='More than 3 executions were triggered')
                     break
                 time.sleep(0.1)
     except Exception as e:
         if 'timeout' not in str(e):
-            print(str(e))
+            sys.stderr.write('%s\n' % str(e))
         env.assertContains('timeout', str(e))
--- a/pytest/test_register.py
+++ b/pytest/test_register.py
@@ -2061,7 +2061,7 @@ def testStreamReaderInfinitLoopOnTrimmingDisabled(env):
 import time
 GB('StreamReader').foreach(lambda x: execute('incr', 'x')).register(batch=2, onFailedPolicy='retry', trimStream=False)
     '''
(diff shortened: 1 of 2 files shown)


pytest/test_register.py
```diff
@@ -2054,3 +2054,54 @@ def testStreamReaderPauseFailedRegistration(env):
...
                 break
             time.sleep(0.1)
+
+@gearsTest(skipOnCluster=True)
+def testStreamReaderInfinitLoopOnTrimmingDisabled(env):
+    script = '''
+import time
+GB('StreamReader').foreach(lambda x: execute('incr', 'x')).register(batch=2, onFailedPolicy='retry', trimStream=False)
+    '''
+    env.expect('rg.pyexecute', script, 'ID', 'test').ok()
```
chayim, line 2064: might as well make this caps for consistency

pytest/test_register.py
```diff
@@ -2054,3 +2054,54 @@ def testStreamReaderPauseFailedRegistration(env):
...
+    env.cmd('xadd', 's', '*', 'foo', 'bar')
+
+    with TimeLimit(5, env, 'Failed waiting for registration to pause'):
+        while True:
+            registrations = env.cmd('RG.DUMPREGISTRATIONS')
+            if registrations[0][7][5] == 2:
+                break
+            time.sleep(0.1)
+
+    env.expect('rg.pyexecute', script, 'ID', 'test', 'UPGRADE').ok()
```
chayim, line 2077: same

pytest/test_register.py
```diff
@@ -2054,3 +2054,54 @@ def testStreamReaderPauseFailedRegistration(env):
...
+        with TimeLimit(1):
+            while True:
+                registrations = env.cmd('RG.DUMPREGISTRATIONS')
+                if registrations[0][7][3] > 2:
+                    self.env.assertTrue(False, message='More than 2 executions was triggered')
+                    break
+                time.sleep(0.1)
+    except Exception as e:
+        if 'timeout' not in str(e):
+            print(str(e))
```
chayim, line 2090: maybe instead a sys.stderr.write

pytest/test_register.py
```diff
@@ -2054,3 +2054,54 @@ def testStreamReaderPauseFailedRegistration(env):
...
+
+    env.cmd('xadd', 's', '*', 'foo', 'bar')
+    env.cmd('xadd', 's', '*', 'foo', 'bar')
+
+    try:
+        with TimeLimit(1):
+            while True:
+                registrations = env.cmd('RG.DUMPREGISTRATIONS')
+                if registrations[0][7][3] != 3:
+                    self.env.assertTrue(False, message='More than 2 executions was triggered')
```
chayim, line 2101:
```
self.env.assertTrue(False, message='More than 2 executions were triggered')
```

pytest/test_register.py
```diff
@@ -2054,3 +2054,54 @@ def testStreamReaderPauseFailedRegistration(env):
...
+                    break
+                time.sleep(0.1)
+    except Exception as e:
+        if 'timeout' not in str(e):
+            print(str(e))
```
chayim, line 2106: sys.stderr.write (remember to add the newline)

chayim (approved):

PR #810, opened 2022-08-16
Fix MOD82576, fix invalid memory access when checking if a trigger is already register

MeirShpilraien:
The invalid memory access happened because each registration was examine even if its not a command reader. This causes wrong casting of structs and in rare cases invalid memory access and crash.

### DIFF
--- a/pytest/test_sessions.py
+++ b/pytest/test_sessions.py
@@ -200,7 +200,7 @@ GB('CommandReader').count().map(lambda x: 'test2 reports %s shards' % str(x)).re
 
 @gearsTest(skipOnCluster=True)
 def testMod4054(env):
-    script = '''
+    script = """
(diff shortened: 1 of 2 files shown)


pytest/test_sessions.py
```diff
@@ -198,6 +198,30 @@ def testCommandReaderRegisterSameTrigger(env):
     '''
     env.expect('RG.PYEXECUTE', script, 'ID', 'test', 'DESCRIPTION', 'desc').error().contains('trigger already registered in this session')
 
+@gearsTest(skipOnCluster=True)
+def testMod4054(env):
+    script = '''
```
chayim, line 203: It's small, and silly - but I'd triple quote with double qutoes (use " in lines 203/222)

chayim (approved):

PR #834, opened 2022-11-29
Added support for ipv6 on clusterset command.

MeirShpilraien:


pytest/test_network.py
```diff
@@ -477,107 +506,112 @@ def testClusterRefreshOnOnlySingleNode(env):
...
+                            'NO-USED',
+                            'password@%s:6379' % shardMock.host,
+                            'NO-USED',
+                            'NO-USED',
+                            '2',
+                            'NO-USED',
+                            '8193',
+                            '16383',
+                            'NO-USED',
+                            'password@%s:%d' % (shardMock.host, shardMock.port)
```
chayim, line 616: What happens with :: and :0: as hosts, both of which are technically valid ipv6?

MeirShpilraien: The code searched for the last `:` and assume that only after comes the port so I do not see why it would be a problem.

chayim (approved):

PR #848, opened 2023-02-19
Redis version bump

chayim:


MeirShpilraien:
@chayim would like your advice if we can put this value somewhere global, currently we have it all around:
https://github.com/RedisGears/RedisGears/blob/b77178c73be4a696079521ab54c1b596badaf8a9/.github/workflows/centos_7.yml#L47
https://github.com/RedisGears/RedisGears/blob/b77178c73be4a696079521ab54c1b596badaf8a9/.github/workflows/centos_8.yml#L33
https://github.com/RedisGears/RedisGears/blob/b77178c73be4a696079521ab54c1b596badaf8a9/.github/workflows/macos_latest.yml#L27
https://github.com/RedisGears/RedisGears/blob/b77178c73be4a696079521ab54c1b596badaf8a9/.github/workflows/ubuntu_bionic.yml#L32
https://github.com/RedisGears/RedisGears/blob/b77178c73be4a696079521ab54c1b596badaf8a9/.github/workflows/ubuntu_focal.yml#L22

chayim:
@MeirShpilraien Not easily **unless** you embrace reusable jobs - which doesn't really work well here. What about a script - and then pulling the artifact from redis-stack dependencies like https://github.com/RedisLabsModules/redis-state-machine/blob/4bc837870abd8256ec807e867ff5bf01af79badc/Makefile#L42

MeirShpilraien:
Closing as this will move us to Redis 7.2RC1

PR #863, opened 2023-03-06
Attempt to fix the flaky macOS tests.

iddm:
Increases the timeout for a failing test and waits on replicas.

### DIFF
--- a/.github/workflows/msrv.yml
+++ b/.github/workflows/msrv.yml
@@ -12,33 +12,39 @@ env:
 
 jobs:
   build:
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-latest
+    container: ubuntu:bionic
 
     steps:
+    - name: Update apt
+      run: apt update
+    - name: Install pre-requisites
+      run: |
+        apt -yqq install software-properties-common
+        add-apt-repository ppa:deadsnakes/ppa
+        add-apt-repository ppa:ubuntu-toolchain-r/test
+        apt update
+        apt -yqq install git gcc-9 g++-9 curl wget python3.10 python3.10-dev dialog subversion chrpath libssl-dev python3-dev clang autoconf automake libtool libssl-dev pkg-config
+        update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-9 9
+        update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-9 9
+        update-alternatives --install /usr/bin/python3 python3 /usr/bin/python3.10 10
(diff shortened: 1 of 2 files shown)


pytests/common.py
```diff
@@ -161,13 +161,19 @@ def test_func():
...
+                            except KeyError:
+                                pass
                             time.sleep(0.1)
             version = env.cmd('info', 'server')['redis_version']
             if skipOnRedis6 and '6.0' in version:
                 env.skip()
             if test_function.__doc__ is not None:
+                if withReplicas:
+                    num_replicas = shardsCount - 1
+                    env.expect('WAIT', num_replicas, 0).equal(num_replicas if decodeResponses else num_replicas.to_bytes())
```
MeirShpilraien, line 176: The wait should come after the `function load` and we should do the wait on each shard like we do here:
https://github.com/RedisGears/RedisGears/pull/863/files#diff-1cf27a9d4d71117ed40ffa9de5085d314d7782a78e8c8620fdd4f6c7cf578511R160

MeirShpilraien: Notice that the number of shards should not be used on the wait command, we always have at most one replica per shard.

pytests/common.py
```diff
@@ -157,18 +157,25 @@ def test_func():
...
+                    return replication_info['slave0']['state']
+
                 for con in shardsConnections(env):
-                    with TimeLimit(10, env, "Failed waiting for replica to be in sync"):
-                        while True:
-                            replication_info = con.execute_command('info', 'replication')
-                            if replication_info['slave0']['state'] == 'online':
-                                break
-                            time.sleep(0.1)
+                    runUntil(env, 'online', synchronise_replicas, timeout=10)
```
MeirShpilraien, line 165: :+1:

pytests/test_errors.py
```diff
@@ -112,7 +112,7 @@ def testJSStackOverflowOnLoading(env):
 test(1);
 redis.register_function('test', test);
     """
-    env.expect('CONFIG', 'SET', 'redisgears_2.lock-redis-timeout', '10000')
+    env.expect('CONFIG', 'SET', 'redisgears_2.lock-redis-timeout', '50000')
```
MeirShpilraien, line 115: I see that we already set a 10 seconds timeout here? So maybe I was wrong and its not timeout?

MeirShpilraien: Ohh I believe the failure was on `testJSStackOverflow` no?
Let add timeout there?

MeirShpilraien: Ok so I have no idea why this is happening, 10 seconds should be more then enough

PR #875, opened 2023-03-31
JavaScript API versioning implementation

iddm:
Implements [the corresponding RFC](https://github.com/RedisGears/RedisGears/pull/873).

It is mandatory to specify the API version in the prologue section using the `api_version` property.

A user may specify a version `X`, but the implementation may decide to pick a more recent API version which is backwards compatible with `X`: a user may specify version 1.0, but the implementation will, for example, provide version 1.1.

When a version is deprecated, it is written so in redis logs, but for the redis client, it is not printed, and the `OK` result is returned. Such functions are stored within the RDB storage, and once loaded, they undergo the same procedure and so the warnings will be printed out again.

This also provides more descriptive prologue parsing errors for the user.

### DIFF
--- a/redisgears_core/src/lib.rs
+++ b/redisgears_core/src/lib.rs
@@ -136,6 +136,7 @@ fn check_redis_version_compatible(ctx: &Context) -> Result<(), String> {
 }
 
 /// The meta information about the gears library instance at runtime.
+#[derive(Debug, Clone, Eq, PartialEq, Ord, PartialOrd, Hash)]
 pub struct GearsLibraryMetaData {
     name: String,
     engine: String,
--- a/.github/workflows/CENTOS_7_REUSABLE.yml
+++ b/.github/workflows/CENTOS_7_REUSABLE.yml
@@ -64,7 +64,7 @@ jobs:
     - name: Install redis
       run: |
         set -x
-        . scl_source enable  devtoolset-9 || true
+        . scl_source enable devtoolset-9 || true
         . scl_source enable llvm-toolset-7 || true
         git clone https://github.com/redis/redis
         cd redis
@@ -78,12 +78,15 @@ jobs:
       run: $HOME/.cargo/bin/cargo fmt -- --check
     - name: Build
       run: |
-        . scl_source enable  devtoolset-9 || true
+        . scl_source enable devtoolset-9 || true
         . scl_source enable llvm-toolset-7 || true
         $HOME/.cargo/bin/cargo build -vv
         $HOME/.cargo/bin/cargo build --release -vv
     - name: Unit tests
-      run: $HOME/.cargo/bin/cargo test --verbose
+      run: |
+        . scl_source enable devtoolset-9 || true
+        . scl_source enable llvm-toolset-7 || true
+        $HOME/.cargo/bin/cargo test --verbose
     - name: Integration tests
       run: |
         PATH=/opt/rh/rh-python38/root/usr/local/bin:$PATH
--- a/.github/workflows/CENTOS_8_REUSABLE.yml
+++ b/.github/workflows/CENTOS_8_REUSABLE.yml
@@ -68,7 +68,9 @@ jobs:
         $HOME/.cargo/bin/cargo build -vv
         $HOME/.cargo/bin/cargo build --release -vv
     - name: Unit tests
-      run: $HOME/.cargo/bin/cargo test --verbose
+      run: |
+        . /opt/rh/gcc-toolset-11/enable
+        $HOME/.cargo/bin/cargo test --verbose
     - name: Integration tests
       run: |
         cd pytests;./run_tests.sh


redisgears_core/src/function_load_command.rs
```diff
@@ -132,17 +99,22 @@ pub(crate) fn function_load_intrernal(
...
+    let lib_ctx = backend.compile_library(
+        &meta_data.name,
+        code,
+        meta_data.api_version,
+        meta_data.config.as_ref(),
+        Box::new(compile_lib_ctx),
+    );
     let lib_ctx = match lib_ctx {
-        Err(e) => return Err(format!("Failed library compilation {}", e.get_msg())),
+        Err(e) => return Err(format!("Failed library compilation:\n{}", e.get_msg())),
```
MeirShpilraien, line 110: I believe we should avoid adding `\n` here, this error will pass the the user eventually and RESP does not support new lines on errors anyway and it could brake the protocol (`redismodule-rs` will protect us from this and will replace the `\n` with space IIRC)

redisgears_plugin_api/src/redisgears_plugin_api/backend_ctx.rs
```diff
@@ -28,6 +30,12 @@ pub enum LibraryFatalFailurePolicy {
     Kill = 1,
 }
 
+impl Default for LibraryFatalFailurePolicy {
+    fn default() -> Self {
+        Self::Abort
```
MeirShpilraien, line 35: Why do we need to implement default?

redisgears_plugin_api/src/redisgears_plugin_api/mod.rs
```diff
@@ -48,6 +49,12 @@ impl GearsApiError {
     }
 }
 
+impl std::fmt::Display for GearsApiError {
```
MeirShpilraien, line 52: :+1:

redisgears_plugin_api/src/redisgears_plugin_api/prologue.rs
```diff
@@ -0,0 +1,349 @@
+/*
+ * Copyright Redis Ltd. 2018 - present
+ * Licensed under your choice of the Redis Source Available License 2.0 (RSALv2) or
+ * the Server Side Public License v1 (SSPLv1).
+ */
+//! The validation mechanism of the user modules.
+//!
+//! Performs the basic validation and API compatibility checks.
+
+use std::collections::HashMap;
```
MeirShpilraien, line 10: Big :+1: 
My only comment here is why not to put this unit on `redisgears_core`?

redisgears_v8_plugin/src/v8_backend.rs
```diff
@@ -52,15 +52,30 @@ unsafe impl GlobalAlloc for Globals {
...
 }
 
 pub(crate) fn get_fatal_failure_policy() -> LibraryFatalFailurePolicy {
-    unsafe { (GLOBAL.backend_ctx.as_ref().unwrap().get_on_oom_policy)() }
+    #[cfg(not(feature = "test"))]
+    unsafe {
+        (GLOBAL.backend_ctx.as_ref().unwrap().get_on_oom_policy)()
+    }
+    #[cfg(feature = "test")]
+    ..Default::default()
```
MeirShpilraien, line 69: Can we return Abort here instead of implement default? I am afraid that by implement default we will accidentally take default value instead of the value set on the configuration.

redisgears_v8_plugin/src/v8_backend.rs
```diff
@@ -52,15 +52,30 @@ unsafe impl GlobalAlloc for Globals {
...
 }
 
 pub(crate) fn gil_lock_timeout() -> u128 {
-    unsafe { (GLOBAL.backend_ctx.as_ref().unwrap().get_lock_timeout)() }
+    #[cfg(not(feature = "test"))]
+    unsafe {
+        (GLOBAL.backend_ctx.as_ref().unwrap().get_lock_timeout)()
+    }
+    #[cfg(feature = "test")]
+    ..Default::default()
```
MeirShpilraien, line 78: Same comment as above

MeirShpilraien (commented):
Implementation look good :+1: 
I have some general comments:
1. API version is not necessarily applies to the Redis object only. We also need to consider the client object, background client, and in the future maybe more. I believe its OK not to handle those now but we will probably need to handle it sooner than we think.
2. The first point lead me to the second point, does API version is on the engine level or on RedisGears level? In other words, does all engine have the same API version or does API versioning need to be handled on each engine separately. I believe we can also answer this question later on (because we only support single engine currently).
3. Tiny comment, I would like to have a test that verify that the `api_version` field is mandatory.

In general looks very good :+1: some tiny comments and some future thoughts ...

MeirShpilraien:
> I think the last time we talked, we strictly spoke about the JS API versioning, so nothing else, and so not RedisGears. Also, the RFC was about the JS API versioning only.

Ok, sounds good, lets keep it JS only for now.

> There are unit and component tests which make sure that the field is mandatory. [Please check this out](https://github.com/RedisGears/RedisGears/pull/875/files#diff-00c29c59efeedc97f3ee9f4e0650cee6935ffb86cf535f3d4bec3d4e09c0230cR283-R288). Also, you may observe that the component tests now all have the module version set up as it has become mandatory.

Yes but I believe a simple end to end test (using our pytests) can not harm and will check the feature end to end.

redisgears_core/src/function_load_command.rs
MeirShpilraien: Please notice that on resp3 we have error blob type that can contain `\n`. We can leverage that for more complex errors like JS traces for example ...

pytests/test_errors.py
```diff
@@ -358,6 +358,12 @@ def testMalformedLibraryMetaData6(env):
     code = '#!js api_version=1.0 name=foo x' # invalid syntax
     env.expect('RG.FUNCTION', 'LOAD', code).error().contains('Invalid or missing prologue.')
 
+@gearsTest()
+def testMalformedLibraryMetaData7(env):
+    code = '#!js name=foo' # no shebang(#!)
```
MeirShpilraien, line 363:
```suggestion
    code = '#!js name=foo' # no API version
```

redisgears_v8_plugin/src/v8_backend.rs
```diff
@@ -286,23 +285,17 @@ impl BackendCtxInterfaceInitialised for V8Backend {
...
-                        .map(|(index, error)| format!("\t{index}. {}", error.get_msg()))
-                        .collect::<Vec<String>>()
-                        .join("\n");
+                let messages = api_version_supported
+                    .validate_code(code)
+                    .iter()
+                    .enumerate()
+                    .map(|(index, error)| format!("\t{}. {}", index + 1, error.get_msg()))
+                    .collect::<Vec<String>>()
+                    .join("\n");
```
MeirShpilraien, line 296: I do not think its a good idea to log with `\n` it will work but will not look so good on the Redis log file. Maybe we can log each entry separately?

redisgears_v8_plugin/src/v8_backend.rs
```diff
@@ -311,6 +304,16 @@ impl BackendCtxInterfaceInitialised for V8Backend {
                     log(compiled_message);
                     script_ctx.compiled_library_api.log(compiled_message);
```
MeirShpilraien, line 305: Why do we need to log here twice? We should always use the `compiled_library_api` for logs so we will get the library name of the log message. the `log` function is just for cases where we want to log general messages and or if we do not have `compiled_library_api` (like if the engine want to log some general message).

MeirShpilraien (commented):
Last small comments

redisgears_v8_plugin/src/v8_backend.rs
```diff
@@ -32,22 +31,6 @@ struct Globals {
     backend_ctx: Option<BackendCtx>,
 }
 
-unsafe impl GlobalAlloc for Globals {
```
MeirShpilraien, line 35: @vityafx I do not think we can drop it without making sure the global allocator is somehow set, otherwise the entire `redisgears_v8_plugin` allocation will not go through Redis allocator and will not be reflected on the stats. On my PR: https://github.com/RedisGears/RedisGears/pull/876 I removed it because I add `redismodule-rs` as a dependency which set the allocator itself.

redisgears_v8_plugin/src/v8_backend.rs
```diff
@@ -287,23 +270,16 @@ impl BackendCtxInterfaceInitialised for V8Backend {
...
-                        messages
-                    );
-
-                    log(compiled_message);
-                    script_ctx.compiled_library_api.log(compiled_message);
-                }
+                    .for_each(|message| {
+                        script_ctx
+                            .compiled_library_api
+                            .log(&format!("Module \"{module_name}\": {message}"))
```
MeirShpilraien, line 281: If module_name is the library name then I believe `compiled_library_api.log` will add it automatically no?

redisgears_v8_plugin/src/v8_backend.rs
MeirShpilraien: Yes I agree, but the user should be somehow able to provide the allocator for `redisgears_v8_plugin` no?

MeirShpilraien: It is today already: https://github.com/RedisGears/RedisGears/blob/34ede8eb1a554fe0cc8efe0c3d052b1ea2789c29/redisgears_plugin_api/src/redisgears_plugin_api/backend_ctx.rs#L32
I guess the conclussion is that we should keep it as I removed it on my PR: https://github.com/RedisGears/RedisGears/pull/876

But maybe we should remove it and re-add it later properly?

.github/workflows/UBUNTU_BIONIC_REUSABLE.yml
```diff
@@ -69,8 +69,10 @@ jobs:
       run: V8_VERSION=${{ inputs.v8_version }} V8_UPDATE_HEADERS=${{ inputs.v8_update_headers }} $HOME/.cargo/bin/cargo build --verbose
     - name: Build release
       run: V8_VERSION=${{ inputs.v8_version }} V8_UPDATE_HEADERS=${{ inputs.v8_update_headers }} $HOME/.cargo/bin/cargo build --release --verbose
-    - name: Tests
-      run: cd pytests;./run_tests.sh
+    - name: Unit tests
```
MeirShpilraien, line 72: We need this change on the other OS's as well no?

pytests/test_errors.py
```diff
@@ -318,34 +327,57 @@ def testFunctionListWithUnknownOption(env):
...
+
+@gearsTest()
+def testMalformedLibraryMetaData2(env):
+    code = '#!js name' # no api version and no module version, invalid prologue.
+    env.expect('RG.FUNCTION', 'LOAD', code).error().contains('Invalid or missing prologue.')
+
+@gearsTest()
+def testMalformedLibraryMetaData3(env):
+    code = '#!js api_version=1.0 foo=bar' # unknown property
+    env.expect('RG.FUNCTION', 'LOAD', code).error().contains('The module name is missing from the prologue.')
```
MeirShpilraien, line 342: This error should be `The library name is missing from the prologue.`. We call the user code a RedisGears library. this matches Redis function, the code there also called library.

redisgears_core/Cargo.toml
```diff
@@ -6,22 +6,24 @@ license = "Redis Source Available License 2.0 (RSALv2) or the Server Side Public
 rust-version = "1.63"
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+[dev-dependencies]
+redis-module = { git = "https://github.com/RedisLabsModules/redismodule-rs", branch = "master", features = ["experimental-api"] }
```
MeirShpilraien, line 10: Why is this needed?

redisgears_core/src/lib.rs
```diff
@@ -135,9 +136,11 @@ fn check_redis_version_compatible(ctx: &Context) -> Result<(), String> {
 }
 
 /// The meta information about the gears library instance at runtime.
+#[derive(Debug, Clone, Eq, PartialEq, Ord, PartialOrd, Hash)]
```
MeirShpilraien, line 139: I am not sure why we need all those derives

redisgears_v8_plugin/src/v8_backend.rs
```diff
@@ -52,15 +53,30 @@ unsafe impl GlobalAlloc for Globals {
...
+    println!("log message: {msg}");
 }
 
 pub(crate) fn get_fatal_failure_policy() -> LibraryFatalFailurePolicy {
-    unsafe { (GLOBAL.backend_ctx.as_ref().unwrap().get_on_oom_policy)() }
+    #[cfg(not(feature = "test"))]
+    unsafe {
+        (GLOBAL.backend_ctx.as_ref().unwrap().get_on_oom_policy)()
+    }
+    #[cfg(feature = "test")]
```
MeirShpilraien, line 69: Why not `#[cfg(test)]`?

MeirShpilraien (commented):
Last small comments

MeirShpilraien:
Also, forgot that we also need to fix the documentation and the documentation examples.

MeirShpilraien (approved):

redisgears_core/src/lib.rs
MeirShpilraien: OK, I can understand that those make sense `Debug, Clone, Eq, PartialEq` but I am not sure about `Ord, PartialOrd, Hash`. Do you see a need to order library meta data. And if so, not sure the default implementation is the right one. WDYT?

MeirShpilraien: OK, convinced :)
Lets keep it.

PR #878, opened 2023-04-06
Specify own RedisGears allocator.

iddm:
Uses the new API of the redismodule crate to set custom global allocator.


MeirShpilraien (commented):
@vityafx This will basically create 2 allocators right? One for the core and one for plugins. I like this Idea but maybe we can take it one step forward and create an allocator per engine. It will give us more flexibility to show more per engine specific memory stats in the future. WDYT?

PR #896, opened 2023-04-27
RFC for KSN API

MeirShpilraien:
[Rendered](https://github.com/RedisGears/RedisGears/pull/896/files?short_path=5ad52b9#diff-5ad52b926da33590d90719661d104664846f23405a5d794315c95c1631847c3c)

### DIFF
--- a/rfcs/3-ksn-api.md
+++ b/rfcs/3-ksn-api.md
@@ -1,4 +1,4 @@
-- RedisGears Issue: [RedisGears/RedisGears#873]()
+- RedisGears Issue: [RedisGears/RedisGears#873](https://github.com/RedisGears/RedisGears/pull/896)
 
 # RedisGears KSN (key space notification) API
 
--- a/rfcs/3-ksn-api.md
+++ b/rfcs/3-ksn-api.md
@@ -1,0 +1,112 @@
+- RedisGears Issue: [RedisGears/RedisGears#873]()
+
+# RedisGears KSN (key space notification) API
+
+This is an RFC to define the JS user facing API for KSN.
+
+# Motivation
+
+Redis has a module API that allow register for KSN events. All the supported events are listed [here](https://redis.io/docs/manual/keyspace-notifications/). The event provide the following information:
+
+1. Event type (number represent the event type)
+2. Event name (usually the command that called the event but could also be `expired`/`evicted`)
+3. Key name
+
+Inside the KSN callback, it is possible to perform any read operation but **it is forbidden to perform any write operations**. The reason for this is describe in details [here](https://github.com/redis/redis/pull/11199).
+
+In order to still perform a write operations as a reaction to KSN. Redis provides a new API called post notification job. This new API allows (inside the notification callback) to register a job that will be invoked when the following conditions holds:
+
+1. It is safe to perform any write operation.
+2. The job will be called atomically along side the key space notification that triggered it.
+
+This document describe options of exposing the described API to the JS user.
+
+## Option 2 - KSN only
+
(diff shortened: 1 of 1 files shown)


rfcs/3-ksn-api.md
```diff
@@ -0,0 +1,112 @@
...
+Inside the KSN callback, it is possible to perform any read operation but **it is forbidden to perform any write operations**. The reason for this is describe in details [here](https://github.com/redis/redis/pull/11199).
+
+In order to still perform a write operations as a reaction to KSN. Redis provides a new API called post notification job. This new API allows (inside the notification callback) to register a job that will be invoked when the following conditions holds:
+
+1. It is safe to perform any write operation.
+2. The job will be called atomically along side the key space notification that triggered it.
+
+This document describe options of exposing the described API to the JS user.
+
+## Option 2 - KSN only
```
thomascaudron, line 24: option 1

rfcs/3-ksn-api.md
```diff
@@ -0,0 +1,112 @@
...
+    // this callback will be called after all the KSN finished.
+    // it is safe to write and read.
+    // the callback has access to the notification name and key name.
+    // The callback has not access to the data at the time the notification happened.
+    let key_name = data.key;
+    client.call('set', 'x', '1'); // performing write command
+});
+
+// Advance API for advance users
+redis.register_ksn_consumer("consumer", "", function(client, data){
```
K-Jo, line 98: can we make the `ksn` in the function name an optional parameter of the function, so it's the same API?

rfcs/3-ksn-api.md
```diff
@@ -0,0 +1,112 @@
...
+    // This callback will be called directly on KSN.
+    // It is impossible to perform writes on this callback
+    // the callback can register a post notification job
+    // to write data atomically with the notification.
+    client.register_post_notification_job(function(client) {
```
K-Jo, line 103: shouldn't we allow to pass some data here?

thomascaudron:
**Can we split the creation of the trigger from the callbacks?**

```js
const trigger = redis.register_notification_consumer("consumer", "keys:");
const sharedData = {};

// This callback will be called directly on KSN.
// It is impossible to perform writes on this callback
trigger.onChange(client, data) {
    sharedData.propertyA = client.call("hget", data.key, "propertyA");
    sharedData.propertyA = ++;
};

// this callback will be called after all the KSN finished.
// it is safe to write and read.
// the callback has access to the notification name and key name.
// The callback has not access to the data at the time the notification happened.
trigger.postChange(client, data){
    client.call("hset", data.key, "propertyA", sharedData.propertyA);
};
```

It makes it clear when it will be executed without going into the details of the KSN.

**Where will be validate that no writes are happening?**
Can we return an error when the library is being loaded or throw an error when it is executed?

rfcs/3-ksn-api.md
```diff
@@ -0,0 +1,112 @@
...
+
+## Option 4 - Simple API for regular users + advance API for advance users
+
+This option is a combination on 2 and 4. We will implement a simple API for most users which will look like option 2. For advance users we will implement option 4.
+
+Example:
+
+```js
+// Simple API for most users
+redis.register_notifications_consumer("consumer", "", function(client, data){
```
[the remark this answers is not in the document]
MeirShpilraien: You can call it whatever you want, its a function argument.

rfcs/3-ksn-api.md
MeirShpilraien: Its a key space notification indicating key space changed. Just trigger does give you any context of why it has been triggered.

K-Jo: register_crud_trigger?

PR #898, opened 2023-05-03
Adding platforms to CI

chayim:
Help up by https://github.com/redis/redis-py/issues/2746


MeirShpilraien:
@chayim we need to add those to the different CI (on PR, on master, and nightly)

chayim:
Abandonded given our path

PR #965, opened 2023-06-22
Add commands.json

thomascaudron:
Add `commands.json` with the overview of the different commands to be used in RedisInsight, Clients. redis.io and others.

### DIFF
--- a/commands.json
+++ b/commands.json
@@ -1,0 +1,118 @@
+{
+    "TFCALL": {
+        "summary": "Invoke a function with Triggers and Function",
+        "since": "2.0.0",
+        "group": "scripting",
+        "complexity": "Depends on the function that is executed.",
+        "arguments": [
+            {
+                "name": "function_name",
+                "type": "string"
+            },
+            {
+                "name": "number_of_keys",
+                "type": "integer"
+            }
+        ]
+    },
+    "TFCALLASYNC": {
+        "summary": "Invoke a async function with Triggers and Function",
+        "since": "2.0.0",
+        "group": "scripting",
+        "complexity": "Depends on the function that is executed.",
+        "arguments": [
+            {
+                "name": "function_name",
+                "type": "string"
+            },
+            {
+                "name": "number_of_keys",
+                "type": "integer"
+            }
+        ]
+    },
+    "TFUNCTION": {
+        "summary": "A container for tfunction commands",
+        "since": "2.0.0",
+        "group": "scripting",
+        "complexity": "Depends on subcommand.",
+        "subcommands": {
+            "TFUNCTION|LOAD": {
+                "summary": "Load a new library to Triggers and Functions",
+                "since": "2.0.0",
+                "group": "scripting",
+                "complexity": "O(1)",
+                "arguments": [
+                    {
+                        "name": "replace",
+                        "pure-token": "replace",
+                        "display_text": "replace",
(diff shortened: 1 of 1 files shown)


commands.json
```diff
@@ -0,0 +1,144 @@
+{
+    "TFCALL": {
+        "summary": "Invoke a function with Triggers and Function",
```
MeirShpilraien, line 3: `TFCALL` is not related to trigger, it just invoke a function ..

commands.json
```diff
@@ -0,0 +1,144 @@
...
+        "since": "2.0.0",
+        "group": "scripting",
+        "complexity": "Depends on the function that is executed.",
+        "arguments": [
+            {
+                "name": "function",
```
MeirShpilraien, line 9: library.function ?

commands.json
```diff
@@ -0,0 +1,144 @@
...
+            {
+                "name": "arg",
+                "type": "string",
+                "optional": true,
+                "multiple": true
+            }
+        ]
+    },
+    "TFCALLASYNC": {
+        "summary": "Invoke a async function with Triggers and Function",
```
MeirShpilraien, line 32: Same comment as above about triggers

commands.json
```diff
@@ -0,0 +1,144 @@
...
+        "since": "2.0.0",
+        "group": "scripting",
+        "complexity": "Depends on the function that is executed.",
+        "arguments": [
+            {
+                "name": "function",
```
MeirShpilraien, line 38: Same comment as above

commands.json
```diff
@@ -0,0 +1,144 @@
...
+                    },
+                    {
+                        "name": "withcode",
+                        "type": "pure-token",
+                        "display_text": "withcode",
+                        "token": "WITHCODE",
+                        "optional": true
+                    },
+                    {
+                        "name": "verbose",
```
MeirShpilraien, line 127: Can we somehow specify that this can be repeated ?

MeirShpilraien (commented):
Same small comments.

commands.json
```diff
@@ -0,0 +1,144 @@
...
+                "summary": "Load a new library to Triggers and Functions",
+                "since": "2.0.0",
+                "group": "scripting",
+                "complexity": "O(1)",
+                "arguments": [
+                    {
+                        "name": "replace",
+                        "pure-token": "replace",
+                        "display_text": "replace",
+                        "token": "REPLACE",
```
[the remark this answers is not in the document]
thomascaudron: The token is used for autocomplete, upper case can be used as is and lower case is to be replaced by user input.

commands.json
```diff
@@ -0,0 +1,144 @@
...
+                ]
+            },
+            "TFUNCTION|DELETE": {
+                "summary": "Delete a library from Triggers and Functions by name",
+                "since": "2.0.0",
+                "group": "scripting",
+                "complexity": "O(1)",
+                "arguments": [
+                    {
+                        "name": "library-name",
```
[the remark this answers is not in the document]
thomascaudron: That should be good, I took it from the [function-delete.json](https://github.com/redis/redis/blob/unstable/src/commands/function-delete.json)

commands.json
thomascaudron: do we have the same behaviour on another command? I couldn't find a pure-token type that had multiple in the Redis.

PR #968, opened 2023-06-22
Fix circle ci to run on tags

MeirShpilraien:


.circleci/config.yml
```diff
@@ -239,6 +244,18 @@ workflows:
           publish: "true"
           <<: *on-integ-branch
 
+      - build-macos-x64:
+          name: OSX [intel] Integration
```
chayim, line 248: ```          name: OSX [intel] Release ```
^^^ so that we can see the names in CI, otherwise it'll say integration.

.circleci/config.yml
```diff
@@ -239,6 +244,18 @@ workflows:
...
+          context: common
+          <<: *on-tags
+          publish: "true"
+
+      - build-macos-arm64:
+          name: OSX [arm64] Integration
```
chayim, line 254: ```          name: OSX [arm64] Release ```
^^^ so that we can see the names in CI, otherwise it'll say integration.

.circleci/config.yml
```diff
@@ -261,6 +278,15 @@ workflows:
             parameters:
               os_version: ["22.04", "20.04", "18.04"]
 
+      - build-arm:
+          name: ARM Integration <<matrix.os_version>>
```
chayim, line 282: ```          name: ARM Release <<matrix.os_version>>```
^^^ so that we can see the names in CI, otherwise it'll say integration.

chayim (approved):

PR #971, opened 2023-06-27
RHEL9 support

chayim:


MeirShpilraien:
@chayim maybe we should label this PR with the platform label to make sure it is working as expected?

MeirShpilraien:
Done on: https://github.com/RedisGears/RedisGears/pull/1037

PR #985, opened 2023-07-13
Avoid --fail-fast on benchmarks. Include tags triggering

filipecosta90:
With `--fail-fast` mode we stop all benchmarks at the first error. Given the benchmark suite of RedisGears is quite simple I suggest we remove this and let the benchmark tool to try to trigger other benchmarks even in the case of some failure. In that manner we'll at least have partial performance data and logs to analyze what happened and confirm if it's test related or env/tool related.

### DIFF
--- a/.github/workflows/nightly-benchmark.yml
+++ b/.github/workflows/nightly-benchmark.yml
@@ -2,10 +2,18 @@ name: Run nightly CI benchmarks
 
 on:
   push:
-    branches:
-      - master
-      - 2.0
-
+      paths-ignore:
+          - '.circleci/**'
+          - 'docs/**'
+          - '*.md'
+      branches:
+          - main
+          - master
+          - '[2-9].[0-9].[0-9]'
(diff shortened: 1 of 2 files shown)


MeirShpilraien:
Thanks @filipecosta90, can you please update the top comment and explain why this change is needed (the removal of fail-fats)?

filipecosta90:
> Thanks @filipecosta90, can you please update the top comment and explain why this change is needed (the removal of fail-fats)?

done @MeirShpilraien :)

MeirShpilraien:
@filipecosta90 wouldn't this has a potential of hiding issues?

filipecosta90:
> @filipecosta90 wouldn't this has a potential of hiding issues?

Nope. it will still fail the benchmarks. The big difference is that it will try to collect more data...

MeirShpilraien (approved):

filipecosta90:
Closing in favor of #995

PR #996, opened 2023-07-21
Fix crash on benchmark when running with V8 main branch.

MeirShpilraien:
The crash started to happened since we merge this PR: https://github.com/RedisGears/RedisGears/pull/973

The purpose of the PR was to initialise the V8 backend lazy only when needed. A side effect was that the V8 might have been initialises on some other thread which are not the main thread. Unfortunately, the V8 platform must be initialised on the main thread (for some reason that I yet been able to understand) and this causes the crash.

The fix is to perform minimal initialisation steps on load time and leave the heavy V8 initialisation processes for the first time its been used. To achieve this, the PR adds a new `on_load` backend interface to `V8Backend` that will be called on loading time and allow the V8 backend to only initialise the platforms.

Depends on: https://github.com/RedisGears/v8-rs/pull/45

PR #997, opened 2023-07-23
Build change to copy files to a second stage docker

chayim:
This PR changes the redisgears build, so that the docker-based builds, copy the ramp files to a second stage docker. At random, the amazonlinux:2 build would fail, due to docker size, and [this bug](https://github.com/docker/buildx/issues/593).

The fix involves modifying the base generator files to include a new second-stage image. The initial image is now used only for building, and the rampfile is copied, to the same location in the second-stage image. GitHub's standard docker loading, then works properly with the second stage image, due to its size being smaller.

As part of this fix #996 has been merged in to this branch proatively.

[Evidence of a working amazonlinux:2 build](https://github.com/RedisGears/RedisGears/actions/runs/5635596689?pr=997)

### DIFF
--- a/.github/workflows/REUSABLE_DOCKER_BUILDER.yml
+++ b/.github/workflows/REUSABLE_DOCKER_BUILDER.yml
@@ -64,6 +64,15 @@ jobs:
               --v8_update_headers ${{inputs.v8_update_headers}} \
               --msrv "${{inputs.msrv}}"
 
+        # https://github.com/docker/build-push-action/issues/321
+        # https://github.com/docker/buildx/issues/593
+        - name: free up some space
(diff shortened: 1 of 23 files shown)


.github/workflows/REUSABLE_DOCKER_BUILDER.yml
```diff
@@ -89,7 +89,7 @@ jobs:
             docker images
 
         - name: extract
-          if: ${{inputs.publish == true}}
+          #if: ${{inputs.publish == true}}
```
MeirShpilraien, line 92: Not needed

MeirShpilraien (approved):

PR #999, opened 2023-07-24
Separate RDB loading from ordinary library loading.

iddm:


redisgears_v8_plugin/src/v8_backend.rs
```diff
@@ -559,6 +574,11 @@ impl BackendCtxInterfaceInitialised for V8Backend {
                 tensor_obj_template,
                 compiled_library_api,
             ));
+
+            script_ctx
+                .is_being_loaded_from_rdb
+                .store(is_loading_rdb, Ordering::SeqCst);
```
MeirShpilraien, line 580: Shouldn't this be reverted somewhere?

MeirShpilraien (commented):
Maybe we can think of a way in which `get_lock_timeout` on backend_ctx will return the timeout base on whether or not its been loaded from rdb. Then the backend will not have to know about whether its loading rdb or not. WDYT?

MeirShpilraien:
We already have it on Redis global state and we can take it from the context flags:
https://github.com/RedisLabsModules/redismodule-rs/blob/96f814ba45a014648884db35e62a0b71de2fe975/src/context/mod.rs#L1311

The question is how do we pass a context to this function..

pytests/test_basics.py
```diff
@@ -536,6 +539,65 @@ def testLibraryConfigurationPersistAfterLoading(env):
...
+    code = """#!js api_version=1.0 name=lib
+redis.registerFunction("test1", function(){
+    return "GOODJOB";
+});
+
+const d = new Date();
+let time = d.getTime();
+
+while (true) {
+    if (new Date().getTime() - time > 1200) {
```
MeirShpilraien, line 572: Maybe we should set it to a little less then 1200 so it will not be flaky?

pytests/test_basics.py
```diff
@@ -536,6 +539,65 @@ def testLibraryConfigurationPersistAfterLoading(env):
...
+def testLibraryLoadingTimesout(env):
+    code = """#!js api_version=1.0 name=lib
+redis.registerFunction("test1", function(){
+    return redis.config;
+});
+while (true) {
+    // do nothing
+}
+    """
+    TIMEOUT_EXPECTED_MS = 5000
```
MeirShpilraien, line 552: Can we set it to 2000 maybe so that the test will be faster?

redisgears_core/src/function_load_command.rs
```diff
@@ -134,24 +132,27 @@ pub(crate) fn function_load_internal(
...
-    let res = lib_ctx.load_library(&GearsLoadLibraryCtx {
-        ctx,
-        gears_lib_ctx: &mut gears_library,
-    });
+    let res = lib_ctx.load_library(
+        &GearsLoadLibraryCtx {
+            ctx,
+            gears_lib_ctx: &mut gears_library,
+        },
+        is_loading_rdb,
```
MeirShpilraien, line 140: Should we add it to `GearsLoadLibraryCtx` struct?

redisgears_v8_plugin/src/v8_script_ctx.rs
```diff
@@ -70,11 +70,10 @@ impl GilStateCtx {
...
     pub(crate) fn git_lock_duration_ms(&self) -> u128 {
-        let duration = match SystemTime::now().duration_since(self.lock_time) {
-            Ok(d) => d,
-            Err(_) => return 0,
-        };
-        duration.as_millis()
+        SystemTime::now()
+            .duration_since(self.lock_time)
+            .unwrap_or_default()
+            .as_millis()
```
MeirShpilraien, line 76: :+1:

redisgears_v8_plugin/src/v8_script_ctx.rs
```diff
@@ -449,6 +464,11 @@ impl LibraryCtxInterface for V8LibraryCtx {
                 )));
             }
         }
+
+        self.script_ctx
```
MeirShpilraien, line 468: Notice that on line 461 on this file. If we return an error we will not unset the `is_being_loaded_from_rdb` value. Maybe we should create a guard struct for this?

MeirShpilraien (commented):
Looks good, some small comments.

MeirShpilraien (dismissed):
:+1:

pytests/test_basics.py
```diff
@@ -598,6 +598,24 @@ def testLibraryLoadingFromRdbTimesout(env):
...
+    # Test that changing the lock-redis-timeout also increases the
+    # rdb one, if it would be lower.
+    env.expect('CONFIG', 'SET', f'{MODULE_NAME}.lock-redis-timeout', 502).equal("OK")
+    current_rdb_timeout_value = 0
+    def set_value(value):
+        nonlocal current_rdb_timeout_value
+        current_rdb_timeout_value = value[1]
+        return value
+    env.expect('CONFIG', 'GET', f'{MODULE_NAME}.rdb-lock-redis-timeout').apply(set_value)
+    env.assertEqual(int(current_rdb_timeout_value), 502)
```
MeirShpilraien, line 617: I think it can be:
```
env.expect('CONFIG', 'GET', f'{MODULE_NAME}.rdb-lock-redis-timeout').apply(lambda v: v[1]).equal(502)
```

MeirShpilraien (dismissed):
Small comment, not a blocker.

MeirShpilraien (approved):

PR #1032, opened 2023-09-15
Add JavaScript debugging

vityafx:
Adds a possibility to debug JavaScript code. To be able to do that, the code must be uploaded to Redis with the "debug" option. Then, in the RESP3 reply, there will be connection hints reported, describing the ways that one may use in order to connect to the remote debugging server in RedisGears.

The debugging is done in a single thread, piece by piece. In the code, it is expressed by working in the cron job so that if nothing is happening with the user script, the thread isn't blocked, and any other work can be performed. However, if the user script is paused, either due to a breakpoint or for whatever other reason, this thread is blocked, waiting until the user unpauses the script. This also provides an atomicity guarantee so that if a script is paused under debugging, nothing else can touch the database.

For the debugging, the existing V8 Inspector API is used.

### DIFF
--- a/pytests/test_debugging.py
+++ b/pytests/test_debugging.py
@@ -333,16 +333,18 @@ class DebuggerClient:
 
 
 def deploy_script(env: Env):
-    env.expect('TFUNCTION', 'DELETE', 'foo')
-    env.expect('TFUNCTION', 'LOAD', 'debug', SCRIPT).contains('The V8 remote debugging server is waiting for a connection')
+    env.expect('TFUNCTION', 'LOAD', 'REPLACE', 'debug', SCRIPT).contains('The V8 remote debugging server is waiting for a connection')
 
-@gearsTest()
+@gearsTest(debugServerAddress="127.0.0.1:9005")
 def testDeployConnectStartDisconnect(env: Env):
     deploy_script(env)
     client = DebuggerClient(env)
     client.disconnect()
 
-@gearsTest()
+    # Check that Redis still operates fine.
+    env.expect('PING').equal(True)
+
+@gearsTest(debugServerAddress="127.0.0.1:9005")
 def testSetBreakpointAndHitIt(env: Env):
     deploy_script(env)
     client = DebuggerClient(env)
@@ -358,7 +360,10 @@ def testSetBreakpointAndHitIt(env: Env):
 
     client.disconnect()
 
-@gearsTest()
+    # Check that Redis still operates fine.
+    env.expect('PING').equal(True)
+
+@gearsTest(debugServerAddress="127.0.0.1:9005")
 def testSetBreakpointAndRemoveIt(env: Env):
     deploy_script(env)
     client = DebuggerClient(env)
@@ -385,16 +390,21 @@ def testSetBreakpointAndRemoveIt(env: Env):
         # We should not receive, however, any other sorts of error.
         raise(e)
 
+    # Check that Redis still operates fine.
+    env.expect('PING').equal(True)
+
     client.disconnect()
+
+    # Check that Redis still operates fine.
+    env.expect('PING').equal(True)
 
 """
 The debugging is considered disabled, when the debug server
 address argument is not correctly specified.
 """
-@gearsTest(debugServerAddress="")
+@gearsTest()
 def testNotPossibleToDebugWhenDebuggingIsDisabled(env: Env):
-    env.expect('TFUNCTION', 'DELETE', 'foo')
-    env.expect('TFUNCTION', 'LOAD', 'debug', SCRIPT).contains('The V8 remote debugging server is waiting for a connection')
+    env.expect('TFUNCTION', 'LOAD', 'REPLACE', 'debug', SCRIPT).contains('The debug server address was not specified in the configuration.')
 
 def assert_load_for_debug_fails(env: Env, connection):
     try:
@@ -411,12 +421,12 @@ def assert_load_for_debug_fails(env: Env, connection):
     except BaseException:
         env.assertTrue(False, message="Didn't fail when it should have.")
 
-@gearsTest(cluster=True)
+@gearsTest(debugServerAddress="127.0.0.1:9005", cluster=True)
 def testNotPossibleWhenInClusterMode(env: Env, master_connection):
     for shard_connection in shardsConnections(env):
         assert_load_for_debug_fails(env, shard_connection)
 
-@gearsTest(withReplicas=True)
+@gearsTest(debugServerAddress="127.0.0.1:9005", withReplicas=True)
 def testNotPossibleOnReplicas(env: Env):
     slave_connection = env.getSlaveConnection()
     assert_load_for_debug_fails(env, slave_connection)
@@ -425,7 +435,7 @@ def testNotPossibleOnReplicas(env: Env):
 The async functions are ran in the background, while the processing
 happens in cron. Let's see Redis can still work fine.
 """
-@gearsTest()
+@gearsTest(debugServerAddress="127.0.0.1:9005")
 def testDebuggingAsyncFunctionInBackground(env: Env):
     deploy_script(env)
     client = DebuggerClient(env)
@@ -445,7 +455,7 @@ def testDebuggingAsyncFunctionInBackground(env: Env):
     future.equal("OK")
     client.disconnect()
 
-@gearsTest()
+@gearsTest(debugServerAddress="127.0.0.1:9005")
 def testKeySpaceTriggerPaused(env: Env):
     import threading
     import time
@@ -474,7 +484,7 @@ def testKeySpaceTriggerPaused(env: Env):
     env.expect('SET', 'keys:*', 'value').equal(True)
     thread.join()
 
-@gearsTest()
+@gearsTest(debugServerAddress="127.0.0.1:9005")
 def testStreamTriggerPaused(env: Env):
     import threading
     import time
--- a/pytests/test_debugging.py
+++ b/pytests/test_debugging.py
@@ -1,5 +1,5 @@
 from redis import ResponseError
-from common import gearsTest, shardsConnections
+from common import gearsTest, shardsConnections, toDictionary
 from common import Env
 from websockets.sync.client import connect
 import json
@@ -343,6 +343,11 @@ def testDeployConnectStartDisconnect(env: Env):
 
     # Check that Redis still operates fine.
     env.expect('PING').equal(True)
+    # After debugging the library should be deleted automatically from
+    # the database, as the only purpose it was uploaded for was to have
+    # a debugging session.
+    env.assertEqual(len(toDictionary(env.cmd('TFUNCTION', 'list', 'library', 'lib', 'v'))), 0)
+
 
 @gearsTest(debugServerAddress="127.0.0.1:9005")
 def testSetBreakpointAndHitIt(env: Env):
--- a/pytests/test_debugging.py
+++ b/pytests/test_debugging.py
@@ -1,0 +1,20 @@
+from common import gearsTest
+from common import toDictionary
+from common import runUntil
+from redis import Redis
+from websockets.client import ClientConnection
+import time
+
+MODULE_NAME = "redisgears_2"
+
+@gearsTest()
+def testDeployConnectStartDisconnect(env):
+    script = '''#!js api_version=1.0 name=foo
+redis.registerFunction("test", function(client) {
+    return 1
+})
+    '''
+
+    env.expect('TFUNCTION', 'DELETE', 'foo')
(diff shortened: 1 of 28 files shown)


MeirShpilraien:
@vityafx I see CI fails. is there any fixes require?

redisgears_core/src/lib.rs
```diff
@@ -868,17 +933,16 @@ fn initialize_v8_backend(
...
         .get_mut(name)
-        .ok_or(RedisError::Str("No such backend"))
+        .ok_or(RedisError::String(format!("No such backend: {name}")))
         .or_else(|_| {
             let uninitialised_backend = get_uninitialised_backends_mut()
                 .remove(name)
-                .ok_or_else(|| RedisError::String(format!("Unknown backend {}", name)))?;
-            let backend = initialize_v8_backend(ctx, uninitialised_backend)?;
+                .ok_or_else(|| RedisError::String(format!("Unknown backend {name}")))?;
+            let backend = initialize_v8_backend(uninitialised_backend)?;
```
MeirShpilraien, line 945: I think this should still get a ctx to indicate that you can only do it when the Redis GIL is held. At least until we will wrap the globals with RedisGILProtected object.

redisgears_core/src/lib.rs
```diff
@@ -1376,7 +1441,7 @@ fn function_debug_command(
         }
         _ => (),
     }
-    let backend = get_backend(ctx, backend_name).map_or(
+    let backend = get_backend(backend_name).map_or(
```
MeirShpilraien, line 1444: Same comment here about the ctx

redisgears_v8_plugin/src/v8_backend.rs
```diff
@@ -113,6 +119,42 @@ fn deny_list() -> &'static HashSet<String> {
...
+    fn log(&self, _: &Record) {}
+    fn flush(&self) {}
+}
+
+struct Logger<'a>(&'a dyn log::Log);
+impl<'a> Logger<'a> {
+    const TARGET_BLACKLIST: [&'static str; 1] = ["tungstenite"];
+}
+
+impl<'a> log::Log for Logger<'a> {
```
MeirShpilraien, line 136: Can you add some comment about this logger please. Explain how it works?

redisgears_v8_plugin/src/v8_backend.rs
```diff
@@ -323,21 +365,188 @@ pub(crate) fn bypass_memory_limit() -> bool {
...
+    Prepared(Option<DebuggerSessionServer>),
+    /// A connection has been established, but the debugging session
+    /// hasn't been started yet.
+    Established(Box<Option<WebSocketServer>>),
+    /// A connection has been established and the debugging session is
+    /// started.
+    Started(Box<ScriptDebuggerSession>),
+}
+
+// TODO remove
```
MeirShpilraien, line 433: So we can remove it?

redisgears_v8_plugin/src/v8_backend.rs
```diff
@@ -493,7 +702,16 @@ impl BackendCtxInterfaceUninitialised for V8Backend {
 
     fn initialize(
         self: Box<Self>,
+        logger: &'static dyn log::Log,
     ) -> Result<Box<dyn BackendCtxInterfaceInitialised>, GearsApiError> {
+        unsafe {
```
MeirShpilraien, line 707: Please explain why this must be unsafe?

redisgears_v8_plugin/src/v8_backend.rs
```diff
@@ -513,6 +735,7 @@ impl BackendCtxInterfaceInitialised for V8Backend {
 
     fn compile_library(
         &mut self,
+        _debug: bool,
```
MeirShpilraien, line 738: So this variable is not used? So why do we need it?

redisgears_v8_plugin/src/v8_backend.rs
```diff
@@ -528,10 +751,11 @@ impl BackendCtxInterfaceInitialised for V8Backend {
         let isolate = V8Isolate::new_with_limits(initial_memory_usage(), initial_memory_limit());
 
         let script_ctx = {
-            let (ctx, script, tensor_obj_template) = {
+            let (ctx, script, tensor_obj_template, inspector) = {
                 let isolate_scope = isolate.enter();
                 let ctx = isolate_scope.new_context(None);
                 let ctx_scope = ctx.enter(&isolate_scope);
+                let inspector = Inspector::new(&ctx_scope);
```
MeirShpilraien, line 758: Maybe we should only create inspector base on the `_debug` variable?

MeirShpilraien: We already have the debugger as Option so why not?

MeirShpilraien (commented):
Thanks Victor, added some comments.
Some general comments I have:
1. I think we should improve the tests, this is a big feature and without massive tests I believe it will be easily break in the future.
2. I do not see where we configure to enable debugger and where it is disabled by default. We should make sure user explicitly enabled the debugger using a specific configuration.

pytests/test_debugging.py
```diff
@@ -0,0 +1,20 @@
...
+def testDeployConnectStartDisconnect(env):
+    script = '''#!js api_version=1.0 name=foo
+redis.registerFunction("test", function(client) {
+    return 1
+})
+    '''
+
+    env.expect('TFUNCTION', 'DELETE', 'foo')
+    env.expect('TFUNCTION', 'LOAD', 'debug', script).equal('OK')
+    client = ClientConnection("ws://127.0.0.1:9005")
```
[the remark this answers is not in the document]
MeirShpilraien: We can make another PR to this branch (instead of master) so it will be easier to review and we will still be able to merge this once its feature complete and includes all the tests.

redisgears_plugin_api/src/redisgears_plugin_api/backend_ctx.rs
```diff
@@ -36,6 +39,50 @@ pub enum LibraryFatalFailurePolicy {
...
 }
 
+/// A payload a [DebuggerBackend] can receive, work with and store.
+/// This type of payload is useful for various backend implementations,
+/// so that those can downcast to their known type, so that that they
+/// can work with it.
+pub type DebuggerBackendPayload = Arc<dyn Any + Send + Sync>;
+/// A type for the script deletion callback. Should be invoked when
+/// the debugger no longer needs the script.
+pub type DebuggerDeleteCallback = Box<dyn FnOnce() -> GearsApiResult>;
```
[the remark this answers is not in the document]
MeirShpilraien: I do not think it is that important to delete it because we can resend it with the `REPLACE` argument.

redisgears_core/src/lib.rs
MeirShpilraien: Isn't it just makes a simple thing for complicated? And for what reason if eventually the right solution is to wrap the globals which are GIL protected with `RedisGILGuard`?

redisgears_v8_plugin/src/v8_backend.rs
MeirShpilraien: Maybe just explain why we need it and why we need the `NoopLogger`

redisgears_v8_plugin/src/v8_backend.rs
MeirShpilraien: Can you explain why we need this logger from the first place? I thought the logger should just work and that inherit it from RedisGears core? Its not the case?

MeirShpilraien: Can we comment it in the code so we will have it documented?

pytests/test_debugging.py
```diff
@@ -0,0 +1,504 @@
...
+
+    """
+    Debugger: step over the next instruction.
+    """
+    def step_over(self) -> [dict]:
+        return self.send_message(DebuggerClientMessage.DEBUGGER_STEP_OVER)
+
+
+def deploy_script(env: Env):
+    env.expect('TFUNCTION', 'DELETE', 'foo')
```
MeirShpilraien, line 336: We use `expect` if we also want to check the reply. If we do not care about the reply probably its better to use `cmd` or `execute_command`

pytests/test_debugging.py
```diff
@@ -0,0 +1,504 @@
...
+    env.expect('TFUNCTION', 'LOAD', 'debug', SCRIPT).contains('The V8 remote debugging server is waiting for a connection')
+
+@gearsTest()
+def testDeployConnectStartDisconnect(env: Env):
+    deploy_script(env)
+    client = DebuggerClient(env)
+    client.disconnect()
```
MeirShpilraien, line 343: Should we check something after? Can you remind me what is expected here?

pytests/test_debugging.py
```diff
@@ -0,0 +1,504 @@
...
+    # Now proceed from the very first instruction till the end.
+    env.assertEqual({}, client.resume()[0]["result"])
+    # The next message should be a "resumed" event from the server.
+    resumed_event = client.receive_event()
+    env.assertEqual("Debugger.resumed", resumed_event["method"])
+    # Now wait for the possible pause event to happen. It should not
+    # happen now, for we have no breakpoints set, and the script
+    # execution should just continue with no problem. We wait for
+    # five seconds to let *any* sort of execution to continue on any VM,
+    # so to avoid the flakiness of the test result.
```
MeirShpilraien, line 377: Maybe we should also do a simple `ping` command and verify that redis is responsive?

pytests/test_debugging.py
```diff
@@ -0,0 +1,504 @@
...
+    client.disconnect()
+
+"""
+The debugging is considered disabled, when the debug server
+address argument is not correctly specified.
+"""
+@gearsTest(debugServerAddress="")
+def testNotPossibleToDebugWhenDebuggingIsDisabled(env: Env):
+    env.expect('TFUNCTION', 'DELETE', 'foo')
+    env.expect('TFUNCTION', 'LOAD', 'debug', SCRIPT).contains('The V8 remote debugging server is waiting for a connection')
```
MeirShpilraien, line 397: Shouldn't we expect here a message that indicating that debugger is disabled?

MeirShpilraien (commented):
Thanks @vityafx ,
Added a few last comments only about the tests.

pytests/test_debugging.py
MeirShpilraien: Also the default should be disable, so how come the other tests does not need to enable it?

pytests/test_debugging.py
vityafx: This requires three more lines to catch the error that the library isn't there. Are you sure you want readability with (in my opinion) excessive try/catch instead of the `expect`, which simply ignores the error (it presumes that the user will check for it), which we don't need? We don't care if the library is there.
I just removed this line for good in favour of using the `REPLACE` flag of the script load instead.

pytests/test_debugging.py
vityafx: Okay, I am sending a PING and check that a reply is sent back. Is this sufficient? Do you suggest to check for anything else? We expect that the debugging session ends well: no crashes, no hangs, Redis can operate fine after.

pytests/test_debugging.py
vityafx: Added a ping check.

pytests/test_debugging.py
vityafx: Guess, my eyes were dry already, indeed a huge overlook. Let me see.

pytests/test_debugging.py
MeirShpilraien: So why do we even need to delete it? Why not just use the replace option on the `TFUNCTION LOAD` command?

pytests/test_debugging.py
MeirShpilraien: For example, do we expect the library to still remain on the server? If so lets check it and if not lets also check it?

pytests/test_debugging.py
vityafx: I am already doing it, as mentioned.

pytests/test_debugging.py
vityafx: We can do that, okay.

MeirShpilraien (dismissed):
There are 2 last comments remain but those are not critical.
We can merge :+1:

pytests/test_debugging.py
MeirShpilraien: Ohh great

pytests/test_debugging.py
vityafx: Checking now that the library is automatically deleted after we disconnect.

MeirShpilraien (approved):

PR #1058, opened 2023-11-21
MOD-6016 link to openssl1.1 on Amazon Linux 2.

MeirShpilraien:


dockerbuilds/dockerfile.rhel.tmpl
```diff
@@ -14,11 +14,14 @@ RUN yum -yqq install rh-python38-python rh-python38-python-devel rh-python38-pyt
...
 RUN yum-config-manager --enable epel
-RUN yum install -qqy git build-essential autoconf libtool curl libssl-dev pkg-config clang wget python3 python3-pip python3-devel lsb-release redhat-lsb-core openssl openssl-devel unzip zip
+RUN yum install -qqy git build-essential autoconf libtool curl libssl-dev pkg-config clang wget python3 python3-pip python3-devel lsb-release redhat-lsb-core openssl11 openssl11-devel unzip zip
 RUN yum install -qqy http://mirror.centos.org/centos/7/extras/x86_64/Packages/centos-release-scl-rh-2-3.el7.centos.noarch.rpm
 RUN yum install -qqy http://mirror.centos.org/centos/7/os/x86_64/Packages/libgfortran5-8.3.1-2.1.1.el7.i686.rpm
 RUN yum install -qqy devtoolset-9 llvm-toolset-7
 RUN yum -yqq install rh-python38-python rh-python38-python-devel rh-python38-python-pip
+RUN ln -s /usr/lib64/pkgconfig/libssl11.pc /usr/lib64/pkgconfig/libssl.pc
+RUN ln -s /usr/lib64/pkgconfig/libcrypto11.pc /usr/lib64/pkgconfig/libcrypto.pc
+RUN ln -s /usr/lib64/pkgconfig/openssl11.pc /usr/lib64/pkgconfig/openssl.pc
```
vityafx, line 24: Trying to fix the issue with less pain? :-) Okay :-)

vityafx (approved):

vityafx:
I think if we are going to make such** changes in CI, we should also have a CI step that runs all of these containers we build and runs Redis there with the module to make sure it runs well there. Also, we need to start with LD_LAZY environment variable there to make sure the linker hasn't missed anything we need at runtime.

MeirShpilraien:
> I think if we are going to make **such** changes in docker file, we should also have a CI step that runs all of these containers we build and runs Redis there with the module to make sure it runs well there. Also, we need to start with the `LD_BIND_NOW` environment variable there to make sure the linker hasn't missed anything we need at runtime.

@vityafx I did run all the platform CI (notice that I tag the PR with `build-all` label). Do you think we need to extend it?

PR #1062, opened 2023-12-06
Fix python tests.

iddm:


pytests/run_tests.sh
```diff
@@ -1 +1,23 @@
...
+
+set -e
+
+FULL_PATH_TO_SCRIPT="$(realpath "${BASH_SOURCE:-$0}")"
+SCRIPT_DIRECTORY="$(dirname "$FULL_PATH_TO_SCRIPT")"
+CALLING_DIRECTORY=`pwd`
+
+printf "\e[31m Python will now install the dependencies system-wide.\e[0m\n"
+
+PIP_BREAK_SYSTEM_PACKAGES=1 python3 -m pip install -r $SCRIPT_DIRECTORY/requirements.txt -U
```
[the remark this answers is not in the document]
iddm: We can do that. I think it is subjective to determine what is better here :-) The whole `run_tests.sh` script is useless in my opinion, and we could just also instruct the user to run `RLTest`.

pytests/requirements.txt
```diff
@@ -0,0 +1,3 @@
+redis==5.0.0rc2
```
MeirShpilraien, line 1: Can you remind me whats the issue with 5.0.1?

pytests/test_debugging.py
```diff
@@ -349,10 +349,10 @@ def testDeployConnectStartDisconnect(env: Env):
     env.assertEqual(len(toDictionary(env.cmd('TFUNCTION', 'list', 'library', 'lib', 'v'))), 0)
 
 
-@gearsTest(debugServerAddress="127.0.0.1:9005")
+@gearsTest(debugServerAddress="127.0.0.1:9006")
 def testSetBreakpointAndHitIt(env: Env):
     deploy_script(env)
-    client = DebuggerClient(env)
+    client = DebuggerClient(env, "ws://127.0.0.1:9006")
```
MeirShpilraien, line 355: I do not like this approach, we had this on redismodule-rs and we always forgot to avoid collisions. Can we think of a better way to distribute the ports?

MeirShpilraien (commented):
Few small comments.
I also see that the mac flow fails, I guess that its because on mac, subporcess python library do not use fork, so it fails to serialize the tests.

pytests/requirements.txt
iddm: The tests freeze for me; I presume the TFCALLASYNC freezes. I wasn't into updating the version so much, so I didn't investigate further.

MeirShpilraien: Ok, lets leave it for another PR.

pytests/test_debugging.py
iddm: Fixed.

MeirShpilraien (approved):

PR #1067, opened 2023-12-13
Fix issue where ksn might be ignored

MeirShpilraien:
The issue is that the `detached_ctx_guard` was released before the `_notification_blocker`. This cause the Redis lock to be released before the notification was resumed. If a notification would have arrive in between the time the Redis lock was released to the time the notifications was resumed, we would have lost it.


iddm (approved):

MeirShpilraien:
Cherry-pick to 2.0: https://github.com/RedisGears/RedisGears/pull/1069

PR #1077, opened 2024-01-11
Fix nightly run failure

MeirShpilraien:
Turns out it is not allowed to raise a termination exception on a co-routine, and as of v12.1 V8 asserts it. To fix the issue we catch the timeout error (termination exception) on `block` function and raise a regular none termination exception. This means that the user will now be able to catch timeout error on the co-routine code and handle it. Currently we do not force any timeout on co-routine so this change does not allow the user to do anything he could not yet already do.
successful nightly run: https://github.com/RedisGears/RedisGears/actions/runs/7491877476

### DIFF
--- a/redisgears_v8_plugin/src/v8_native_functions.rs
+++ b/redisgears_v8_plugin/src/v8_native_functions.rs
@@ -300,7 +300,14 @@ pub(crate) fn get_backgrounnd_client<'isolate_scope, 'isolate>(
 
             let _block_guard = ctx_scope.set_private_data(0, &true); // indicate we are blocked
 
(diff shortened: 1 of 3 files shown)


redisgears_v8_plugin/src/v8_native_functions.rs
```diff
@@ -300,7 +300,14 @@ pub(crate) fn get_backgrounnd_client<'isolate_scope, 'isolate>(
 
             let _block_guard = ctx_scope.set_private_data(0, &true); // indicate we are blocked
 
-            Ok(script_ctx_ref.call(&f, ctx_scope, Some(&[&c.to_value()]), GilStatus::Locked))
+            let trycatch = isolate_scope.new_try_catch();
```
chayim, line 303: Thanks for teaching me something!

chayim (approved):

PR #1090, opened 2024-02-29
Fixing OSX builds via venv

chayim:
Lifting the change from #1089, for its master branch equivalent


MeirShpilraien (approved):

MeirShpilraien (dismissed):

MeirShpilraien (approved):

MeirShpilraien:
Jammy did not run because we only change the circle job.

PR #1095, opened 2024-03-17
Downgrade scripts log level to debug on enterprise cluster.

MeirShpilraien:
Logs that was writen from scripts will now be written with debug log level when running on enteprise cluster.

### DIFF
--- a/redisgears_core/src/lib.rs
+++ b/redisgears_core/src/lib.rs
@@ -14,7 +14,7 @@ use keys_notifications_ctx::KeySpaceNotificationsCtx;
 use redis_module::redisvalue::RedisValueKey;
 use redis_module::{
     BlockingCallOptions, CallOptionResp, CallOptionsBuilder, CallResult, ContextFlags, ErrorReply,
-    PromiseCallReply, RedisGILGuard,
+    PromiseCallReply, RedisGILGuard, Version,
 };
 use redisgears_plugin_api::redisgears_plugin_api::backend_ctx::BackendCtxInterfaceInitialised;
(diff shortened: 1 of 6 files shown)


redisgears_core/src/lib.rs
```diff
@@ -938,21 +943,36 @@ fn js_init(ctx: &Context, _args: &[RedisString]) -> Status {
...
+    let is_enterprise = ctx
+        .server_info("server")
+        .field("rlec_version")
+        .map_or(false, |_| true);
+
+    let redis_version = RedisVersion {
+        version,
+        is_enterprise,
+    };
+
```
iddm, line 963: Perhaps, we could inline this code directly into the `get_and_verify_redis_version_compatible` function? And make this function return `RedisVersion` instead of the `Version`?

redisgears_core/src/lib.rs
```diff
@@ -590,7 +589,13 @@ impl DbPolicy {
     }
 }
 
+struct RedisVersion {
```
iddm, line 592: Can we add `#[derive(Debug, Copy, Clone)]` on top?

iddm (approved):

redisgears_core/src/lib.rs
MeirShpilraien: Copy we can not because Version do not derive Copy. But we sure can Clone and Debug.

MeirShpilraien: Actually, Version also do not implement Clone.

iddm (approved):

PR #1104, opened 2024-05-13
Fix nightly failure, avoid calling get_exception_msg until we know an exception was actually occured.

MeirShpilraien:


pytests/test_debugging.py
```diff
@@ -459,6 +459,9 @@ def testNotPossibleOnReplicas(env: Env):
 """
 @gearsTest(debugServerAddress=get_debug_server_address())
 def testDebuggingAsyncFunctionInBackground(env: Env):
+    # Disable for now, the debugging feature is only available on master
+    # and we do not want to fail the CI because of it.
+    env.skip()
```
MeirShpilraien, line 464: @iddm I am disabling this test for now cause it hang. If maybe have time to look it at maybe its an easy fix?

iddm (approved):

MeirShpilraien:
Cherry-pick is not needed as we do not have this issue on 2.0:
https://github.com/RedisGears/RedisGears/blob/54eb1474fa8df61fb4022db53f27fac5986a37ca/redisgears_v8_plugin/src/v8_backend.rs#L576

PR #1106, opened 2024-05-15
Implement the "noop" mode.

iddm:
Adds a "noop" mode where the module acts like it is a fully-operational RedisGears v2 (triggers and functions) module, while in fact does nothing. It still loads the RDB, has the same commands, but the commands do not perform any action, though still can be invoked. The configuration options remain as well.


redisgears_core/src/noop/mod.rs
```diff
@@ -0,0 +1,350 @@
...
+            {
+                flags: [ReadWrite, Access, Update],
+                begin_search: Index({ index : 2}),
+                find_keys: Keynum({ key_num_idx : 0, first_key : 1, key_step : 1 }),
+            }
+        ],
+    }
+)]
+    fn function_call(_ctx: &Context, _args: Vec<RedisString>) -> RedisResult {
+        RedisResult::Ok(RedisValue::Null)
```
MeirShpilraien, line 191: Return some error message

redisgears_core/src/noop/mod.rs
```diff
@@ -0,0 +1,350 @@
...
+            {
+                flags: [ReadWrite, Access, Update],
+                begin_search: Index({ index : 2}),
+                find_keys: Keynum({ key_num_idx : 0, first_key : 1, key_step : 1 }),
+            }
+        ],
+    }
+)]
+    fn function_call_async(_ctx: &Context, _args: Vec<RedisString>) -> RedisResult {
+        RedisResult::Ok(RedisValue::Null)
```
MeirShpilraien, line 209: Error message

redisgears_core/src/noop/mod.rs
```diff
@@ -0,0 +1,350 @@
...
+    {
+        name: "_rg_internals.function",
+        flags: [MayReplicate, DenyScript, NoMandatoryKeys],
+        enterprise_flags: [ProxyFiltered],
+        arity: -3,
+        key_spec: [],
+    }
+)]
+    fn function_command_on_replica(_ctx: &Context, _args: Vec<RedisString>) -> RedisResult {
+        RedisResult::Ok(RedisValue::Null)
```
MeirShpilraien, line 222: This is an internal command between to replica and AOF, it should return a simple OK reply

redisgears_core/src/noop/mod.rs
```diff
@@ -0,0 +1,350 @@
...
+    #[command(
+    {
+        name: "tfunction",
+        flags: [MayReplicate, DenyScript, NoMandatoryKeys],
+        arity: -2,
+        key_spec: [],
+    }
+)]
+    fn function_command(_ctx: &Context, _args: Vec<RedisString>) -> RedisResult {
+        RedisResult::Ok(RedisValue::Null)
```
MeirShpilraien, line 234: Return an error message

redisgears_core/src/noop/mod.rs
```diff
@@ -0,0 +1,350 @@
...
+            {
+                flags: [ReadWrite, Access, Update],
+                begin_search: Index({ index : 3}),
+                find_keys: Range({ last_key: 0, steps: 1, limit: 0 }),
+            }
+        ],
+    }
+)]
+    fn update_stream_last_read_id(_ctx: &Context, _args: Vec<RedisString>) -> RedisResult {
+        RedisResult::Ok(RedisValue::Null)
```
MeirShpilraien, line 253: This is an internal command between to replica and AOF, it should return a simple OK reply

pytests/test_noop.py
```diff
@@ -0,0 +1,50 @@
...
+    env.assertEqual(env.tfcallAsync('gears_example', 'asyncfoo'), None)
+    env.assertEqual(env.cmd('get', 'triggers_counter'), None)
+
+    # Key triggers
+    env.cmd('set', 'key1', 'bar')
+    env.assertEqual(env.cmd('get', 'triggers_counter'), None)
+
+    # Stream triggers
+    env.cmd('xadd', 'stream1', '*', 'foo', 'bar')
+    env.assertEqual(env.cmd('get', 'triggers_counter'), None)
```
MeirShpilraien, line 50: We should also check AOF Loading with `_rg_internals.*` commands

MeirShpilraien (commented):
:+1: 
I believe that we also need to remove the dependency on the ramp.yml so that only the noop module will be added to the final zip.

PR #1108, opened 2024-06-16
End of preview banner

LiorKogan:


jfloodnet:
@LiorKogan  will there be a full announcement coming regarding the fate of "Triggers and functions" and/or "RedisGears"?

LiorKogan:
@jfloodnet There will be no public announcement about the discontinuation of RedisGears 2.0 as it was never GAed.

jfloodnet:
> @jfloodnet There will be no public announcement about the discontinuation of RedisGears 2.0 as it was never GAed.

I appreciate the reply. Apologies for my ignorance, but does that imply that RedisGears 1.0 will continue to receive support?

LiorKogan:
@jfloodnet RedisGears 1.x was GAed. We may release critical patches, but that's about it. We don't plan to invest in it further.

PR #1116, opened 2024-10-13
Remove centos7 build, update version to 1.2.10

MeirShpilraien:


spicy-sauce (approved):
LGTM. RHEL7 reached its EOL.

PR #1133, opened 2025-08-25
MOD-10638 Add event-dispatch workflow to master

galcohen-redislabs:
Adding this workflow to master is needed because of a github workflows limitation: workflows can only be triggered if they exist on the default branch, and we want to trigger this dispatch flow from side branches as well as from the 1.x branches.

### DIFF
--- a/.github/workflows/event-dispatch.yaml
+++ b/.github/workflows/event-dispatch.yaml
@@ -9,8 +9,9 @@ on:
         type: choice
         default: focal
         options:
-          - rhel8
-          - rhel9
+          - "rhel8"
+          - "rhel9"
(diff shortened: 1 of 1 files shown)


.github/workflows/event-dispatch.yaml
```diff
@@ -0,0 +1,41 @@
+name: Gears Manual Run
+
+on:
+  workflow_dispatch:
+    inputs:
+
+      os:
+        description: "Which os to run on. Leave empty to run on all"
```
ephraimfeldblum, line 8: this description doesn't match what is actually occurring in the script

PR #1135, opened 2026-02-25
MOD-13818 Set version to 8.4

galcohen-redislabs:


gabsow:
The workflow matrix only defines x86 runners:

```yaml
matrix:
  platform:
    - os: focal
      arch: x86
      runner: ubuntu-latest
    - os: jammy
      arch: x86
      runner: ubuntu-latest
    - os: rhel8
      arch: x86
      runner: ubuntu-latest
    - os: rhel9
      arch: x86
      runner: ubuntu-latest
```

We're going to need ARM binaries for 8.4. Can you add ARM entries to the matrix (e.g. using `runner: ubuntu-24.04-arm` or a self-hosted ARM runner)? Without that, this workflow will only produce x86 artifacts.

galcohen-redislabs:
Gears doesn't work on arm (not sure why exactly but Meir told me there were several pitfalls and therefore only x86 is supposed)

gabsow (approved):

